PR #1, opened 2024-02-16
add test cases for AllDifferentValue

AllenZzw:
Add a satisfiable and an unsatisfiable simple cases for the AllDifferentValue constraint.

### DIFF
--- a/crates/huub/src/propagator/all_different.rs
+++ b/crates/huub/src/propagator/all_different.rs
@@ -49,3 +49,45 @@ impl Propagator for AllDifferentValue {
 		}
 	}
 }
+
+mod tests {
+    #[test]
+	fn test_all_different_value_sat() {
+		use super::*; 
+		use flatzinc_serde::RangeList;
+		use crate::solver::engine::int_var::IntVar; 
+		use pindakaas::Cnf;
+
+		let mut slv = Cnf::default().into(); 
+		let a = IntVar::new_in(&mut slv, RangeList::from_iter([1..=4])); 
+		let b = IntVar::new_in(&mut slv, RangeList::from_iter([1..=4])); 
+		let c = IntVar::new_in(&mut slv, RangeList::from_iter([1..=4])); 
+
+		slv.add_propagator(Box::new(AllDifferentValue::new(vec![a, b, c])));
+		slv.solve(|val| {
+			assert_ne!(val(a.into()), val(b.into()));
+			assert_ne!(val(b.into()), val(c.into()));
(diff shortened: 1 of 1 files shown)


AllenZzw:
I tried to print more debug message and get the following: 

```
running 1 test
Propagate []
Notify new decision level
notify_assignment: var: Var(7), val: false, persistent: false
        propagate int var 2 event Fixed to propagator 0 with data 2
notify_assignment: var: Var(8), val: true, persistent: false
        propagate int var 2 event Fixed to propagator 0 with data 2
        Propagating: IntView(VarRef(0)) != 2
        Propagating: IntView(VarRef(1)) != 2
        Propagating: IntView(VarRef(0)) != 2
        Propagating: IntView(VarRef(1)) != 2
Propagate [Lit(-2), Lit(-5), Lit(-2), Lit(-5)]
notify_assignment: var: Var(2), val: false, persistent: false
        propagate int var 0 event Fixed to propagator 0 with data 0
notify_assignment: var: Var(1), val: true, persistent: false
        propagate int var 0 event Fixed to propagator 0 with data 0
notify_assignment: var: Var(5), val: false, persistent: false
        propagate int var 1 event Fixed to propagator 0 with data 1
notify_assignment: var: Var(4), val: true, persistent: false
        propagate int var 1 event Fixed to propagator 0 with data 1
Propagate []
Check model with value
thread 'propagator::all_different::tests::test_all_different_value_unsat' panicked at crates/huub/src/propagator/all_different.rs:89:13:
assertion `left != right` failed
  left: Some(Int(1))
 right: Some(Int(1))
```

It seems like the IPASIR-UP will directly call "check_model" when all observed Boolean variables are fixed. This failure can be avoid if have "check_model" implemented for the AllDifferentValue constraint.

AllenZzw:
Hi, 

I have a look at the source code of the CaDiCal solver, and find the following code snippet which calls "cb_check_found_model". 

```
bool Internal::external_check_solution () {
  if (!external->propagator) return true;
  bool trail_changed = true;
  bool added_new_clauses = false;
  while (trail_changed || added_new_clauses) {
    notify_assignments ();
    if (!satisfied ()) break;
    trail_changed = false; // to be on the safe side
    added_new_clauses = false;
    
    ...

    bool is_consistent = external->propagator->cb_check_found_model(etrail);

    if (is_consistent) {
      LOG("Found solution is approved by external propagator.");
      return true;
    }
```

It seems like the CaDiCal would indeed call "notify_assignments" before "cb_check_found_model". Is the observed behavior due to the interface provided by the pindakass library?

PR #24, opened 2024-05-10
Fix/propagation

AllenZzw:
Fix propagation for linear and minimum propagators: 
- fix over-strong explanation for linear propagator 
- fix get_int_lit for the case of linear integer view with negative scale value
- update satisfiable test cases to check all solutions

### DIFF
--- a/crates/huub/src/solver/engine.rs
+++ b/crates/huub/src/solver/engine.rs
@@ -460,7 +460,7 @@ impl ExplainActions for State {
 					if scale.is_positive() {
 						self.int_vars[var].get_bool_lit(LitMeaning::Less(val + 1))
 					} else {
-						self.int_vars[var].get_bool_lit(LitMeaning::GreaterEq(val))
+						self.int_vars[var].get_bool_lit(LitMeaning::GreaterEq(val + 1))
 					}
 				}
 			},
--- a/crates/huub/src/solver/engine.rs
+++ b/crates/huub/src/solver/engine.rs
@@ -392,13 +392,16 @@ impl ExplainActions for State {
 		match var.0 {
 			IntViewInner::VarRef(iv) => self.int_trail[self.int_vars[iv].lower_bound],
 			IntViewInner::Const(i) => i,
-			IntViewInner::Linear { var, scale, offset } => {
-				if scale.is_positive() {
+			IntViewInner::Linear {
+				transformer: transform,
+				var,
+			} => {
+				if transform.positive_scale() {
 					let lb = self.int_trail[self.int_vars[var].lower_bound];
-					IntView::linear_transform(lb, scale, offset)
+					transform.transform(lb)
 				} else {
 					let ub = self.int_trail[self.int_vars[var].upper_bound];
-					IntView::linear_transform(ub, scale, offset)
+					transform.transform(ub)
 				}
 			}
 		}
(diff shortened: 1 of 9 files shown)


crates/huub/src/solver/engine.rs
```diff
@@ -460,7 +460,7 @@ impl ExplainActions for State {
 					if scale.is_positive() {
 						self.int_vars[var].get_bool_lit(LitMeaning::Less(val + 1))
```
Dekker1: I would imagine that this then also has to be `Less(val)` rather than `val + 1`. Looks like a copying mistake

AllenZzw: This is correct. Suppose we have constants a, b > 0, and now we want the literal for [bx < a]. Reasoning as follows: 
- bx < a is equivalent to x < float(a/b)
- For integer division, a/b will become the integer floor(a/b), and x < float(a/b) is equivalent to x <= floor(a/b)
- The literal [x <= floor(a/b)] is equivalent to get the literal for [x < floor(a/b) + 1]
Therefore, we still need this "+ 1" in the case of positive scaling value.

Dekker1: Ah, good to know there is some reasoning behind this, but I don't think this works when `a=1` and `b=1`, so I would imagine that there are other exceptions as well

AllenZzw: Ah, yes. It seems like we need to separate the case where a divides b or not.

Dekker1: I don't think it's just about whether there is a division or not. I think there is a problem with the reasoning. The same problem happens for `a=2` and `b=2`

Dekker1: I think we might need to tell the transformation function whether to round up or down in the transformation (based on whether you are using a `Less` or `GreaterEq` literal)

AllenZzw: Yes, the step "x < float(a/b) is equivalent to x <= floor(a/b)" is incorrect. If a divides b, then `float(a/b)` is equal to `floor(a/b)`, which means `x < float(a/b)` should be equivalent to `x < floor(a/b)` instead of `x <= floor(a/b)`.

Dekker1: Okay, but then that would actually mean we don't need the `+ 1` right?

AllenZzw: It seems to me there are two cases: 
- when `a` divides `b`, `floor(a/b)` is equal to `float(a/b)`, and we don't need the `+ 1`. 
- when `a` does not divide `b`, `x < float(a/b)` is equivalent to `x <= floor(a/b)`, then we need the `+ 1` since now we want the literal `x < floor(a/b) + 1`. 
For the second case, imagine that `a=2` and `b = 3`. We want `x < 1.5` which is `x <= 1` and equivalently `x < 2`.

Dekker1: I think in either case you can say that `x < float(a/b)` is equivalent to `x < floor(a/b)`, because x cannot take a value in the range `floor(a/b)>..float(a/b)`. So I'm fairly certain that you never need the `+ 1`

AllenZzw: The same thing happens for `GreaterEq` literal. Suppose we want `ax >= b`:
- when `a = 2` and `b = 3`, then `x >= 1.5` is equivalent to `x >= 2` and `x >= floor(a/b) + 1` 
- when `a = 2` and `b = 2`, then `x >= 1` is equivalent to `x >= floor(a/b)`

AllenZzw: The argument that "x cannot take a value in the range `floor(a/b)>..float(a/b)`" seems not true to me. `x` can still take the value `floor(a/b)`, right? 
In the case of `a = 2` and `b = 3`, we say `2x < 3`, which means `x < 1.5`. But `x` can still take the value `1`?

AllenZzw: Yes, true. I am thinking how to update the function `rev_linear_transform`.

Dekker1: You are certainly right about that one. That should be `x < 2`. Let's think about this over the weekend and see how we can make this work correctly and consistently

AllenZzw: Yes, I would also take a look at gecode and chuffed and see how they handle these cases.

Dekker1: I've had a look at Chuffed and Gecode, and it seem that it really is that on Less you have to `ceil` and on GreaterEq you have to `floor`.

I have a good idea how to get these changes into a reusable way into the solver. I'll edit this pull request somewhere today when I have time

PR #26, opened 2024-05-13
Feat/reified linear propagator

AllenZzw:
Add "reify" function in linear propagator for half reification


AllenZzw:
@Dekker1 I made some changes in propagation_context.rs. The idea is to register the reason when we assign value to some unsigned Boolean variable. Also, I fix the issue where the reason for assigning a bool value is not equivalent to the detected conflict, as shown in line 29 - 38 in propagation_context.rs. 

To make things work, I need to update the "assign" function in sat trail to allow assigning a value that is currently different from the sat trail. Even though all test cases passed, this change seems to be fundemental. We could have a discussion before merging this change into the develop branch.

PR #35, opened 2024-05-23
[WIP] Fix/jobshop propagation

AllenZzw:
Fix several issues when testing the jobshop scheduling instances

### DIFF
--- a/crates/huub/src/propagator/int_lin_le.rs
+++ b/crates/huub/src/propagator/int_lin_le.rs
@@ -1,6 +1,6 @@
 use pindakaas::Lit as RawLit;
 
-use super::{reason::ReasonBuilder, ExplainActions, InitializationActions, PropagationActions};
+use super::{reason::ReasonBuilder, InitializationActions, PropagationActions};
 use crate::{
 	helpers::opt_field::OptField,
 	propagator::{conflict::Conflict, int_event::IntEvent, Propagator},
@@ -9,7 +9,7 @@ use crate::{
 		value::IntVal,
 		view::{BoolViewInner, IntView, IntViewInner},
 	},
-	BoolView, Conjunction,
+	BoolView,
 };
 
 #[derive(Debug, Clone, PartialEq, Eq, Hash)]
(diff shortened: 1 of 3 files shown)


AllenZzw:
The following instance will output "Assertion failed: (v.level > 0), function protect_reasons, file collect.cpp, line 109." from the CaDiCal solver: 
```
include "globals.mzn";

int: n =10;
int: m =5;
int: span=400;

array [1..n, 1..2*m] of int: job = [| 1,21, 0,53, 4,37, 3,55, 2,34
       | 0,21, 3,21, 4,16, 2,26, 1,71
       | 3,39, 4,16, 1,42, 2,31, 0,12
       | 1,37, 0,55, 4,26, 2,66, 3,77
       | 0,25, 3,34, 2,64, 1,19, 4,37
       | 1,22, 2,43, 4,18, 0,14, 3,13
       | 3,41, 4,32, 1,33, 2,28, 0,17
       | 2,38, 0,31, 1,41, 3,24, 4,33
       | 3,17, 1,14, 4,25, 0,44, 2,15
       | 4,17, 3,15, 2,43, 1,17, 0,17 |];

array [1..n] of int: one = [ 1 | i in 1..n];

int: nn = n; 
array [1..nn, 1..m] of var 0..span: t;
var 0..span: objective;

constraint forall(i in 1..nn, j in 1..m-1) ( t[i,j] + job[i, 2*j] <= t[i, j+1] );

constraint forall(k in 1..m) (
   let {
      array[1..n] of int: d = [ job[i, 2*j] | i in 1..n, j in 1..m where job[i,2*j-1] = k-1],
      array[1..n] of var 0..span: s = [ t[i,j] | i in 1..n, j in 1..m where job[i,2*j-1] = k-1]
   }
   in
   cumulative(s, d, one, 1)
   );

constraint maximum(objective, [ t[i, m] + job[i, 2*m] | i in 1..nn]);

solve :: int_search([t[i,j] | i in 1..nn, j in 1..m], smallest, indomain_min, complete) minimize objective;
```

crates/huub/src/solver/engine.rs
```diff
@@ -409,7 +409,9 @@ impl State {
 			Ok(reason) => {
 				let _ = self.reason_map.insert(lit, reason);
 			}
-			Err(true) => (),
+			Err(true) => {
+				let _ = self.reason_map.insert(lit, Reason::Eager(Box::from([])));
```
Dekker1, line 413: I'm not sure why we would store this. We already allowed propagating literals without reason. Why does the empty reason need to be recorded?

crates/huub/src/propagator/int_lin_le.rs
```diff
@@ -149,35 +149,27 @@ impl<const B: usize> Propagator for IntLinearLessEqBoundsImpl<B> {
...
+						None
+					} else {
+						Some(actions.get_int_lower_bound_lit(*v))
+					}
+				})
+				.collect();
+			if let Some(r) = self.reified.get() {
+				var_lits.push(BoolView(BoolViewInner::Lit(*r)))
+			}
+			let reason = ReasonBuilder::Eager(var_lits);
```
Dekker1, line 168: Looking at it, the eager reason looks the same to me as the lazy explanation (apart from the fact that the bounds might already be stronger in the lazy expalnation). What was the problem with the explanation? Why are we switching to Eager explanation?

Dekker1 (commented):
Is the "Assertion failed: (v.level > 0)" still triggered with these changes?

AllenZzw:
> Is the "Assertion failed: (v.level > 0)" still triggered with these changes?

Yes. It seems like there are some literals with level 0 exist in the trail of CaDiCal.

PR #58, opened 2024-07-09
Edge finding propagator for disjunctive

AllenZzw:
Add edge finding propagator for disjunctive constraint.

### DIFF
--- a/crates/huub/src/model/constraint.rs
+++ b/crates/huub/src/model/constraint.rs
@@ -8,7 +8,7 @@ use pindakaas::{
 use rangelist::RangeList;
 
 use crate::{
-	actions::explanation::ExplanationActions,
+	actions::{explanation::ExplanationActions, inspection::InspectionActions},
 	helpers::{div_ceil, div_floor},
 	model::{
 		bool::BoolView,
@@ -19,6 +19,7 @@ use crate::{
 		all_different_int::AllDifferentIntValue,
 		array_int_minimum::ArrayIntMinimumBounds,
 		array_var_int_element::ArrayVarIntElementBounds,
+		disjunctive::DisjunctiveEdgeFinding,
 		int_lin_le::{IntLinearLessEqBounds, IntLinearLessEqImpBounds},
 		int_lin_ne::{IntLinearNotEqImpValue, IntLinearNotEqValue},
 		int_pow::IntPowBounds,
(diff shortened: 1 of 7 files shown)


AllenZzw:
The details of comparison with the decomposed disjunctive using minizinc-benchmark instances are as follows (1 minute timeout): 

Negative Status Changes:
------------------------
OPTIMAL_SOLUTION -> SATISFIED:
  - jobshop_bench/jobshop.mzn jobshop_bench/abz6.dzn

Timing Changes (>±10.0%):
------------------------
- (-24.0%: 1.1s -> 0.9s) jobshop_bench/jobshop.mzn jobshop_bench/mt06.dzn
- (-40.3%: 6.7s -> 4.0s) jobshop_bench/jobshop.mzn jobshop_bench/la01.dzn
- (-48.3%: 12.8s -> 6.6s) jobshop_bench/jobshop.mzn jobshop_bench/la02.dzn
- (-56.9%: 14.2s -> 6.1s) jobshop_bench/jobshop.mzn jobshop_bench/la03.dzn
- (-67.6%: 12.5s -> 4.1s) jobshop_bench/jobshop.mzn jobshop_bench/la04.dzn
- (-78.8%: 14.8s -> 3.1s) jobshop_bench/jobshop.mzn jobshop_bench/la05.dzn

Objective Changes (>±10.0%):
------------------------
- (89.7%: MIN 1054.00 -> 1999.00) jobshop_bench/jobshop.mzn jobshop_bench/mt10.dzn

AllenZzw:
The propagator still runs into some issue with the following message: 

> thread 'main' panicked at crates/huub/src/solver/engine.rs:764:14:
literals part of explanations should have been created during propagation
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
=====ERROR=====

[jobshop.fzn.json](https://github.com/user-attachments/files/16411684/jobshop.fzn.json)

Dekker1:
That sounds like you're running into the problem we've already discussed. Cadical requires us to propagate the literals in the reason before the literal it causes. This means that we cannot strengthen explanations, because it might need to introduce new Boolean variables.

This is exactly the case you're running into, where you are requesting a literal that doesn't yet exist during explanation.

(Or otherwise, it might be a bug in the implementation)

Dekker1 (approved):

PR #81, opened 2024-08-16
add binding to constant true/false for BoolExpr::And/BoolExpr::Or

AllenZzw:


crates/huub/src/model/bool.rs
```diff
@@ -204,13 +204,21 @@ impl BoolExpr {
...
 						BoolViewInner::Const(false) => {}
 						BoolViewInner::Const(true) => return bind_const(&mut slv.oracle, true),
-						BoolViewInner::Lit(l) => lits.push(Formula::Atom(l)),
+						BoolViewInner::Lit(l) => {
+							lits.push(Formula::Atom(l));
+							all_false = false;
+						}
 					}
 				}
+				if all_false {
```
Dekker1: Can't we just check `lits.is_empty()`?

crates/huub/src/model/bool.rs
```diff
@@ -222,13 +230,20 @@ impl BoolExpr {
...
 						BoolViewInner::Const(true) => {}
 						BoolViewInner::Const(false) => return bind_const(&mut slv.oracle, false),
-						BoolViewInner::Lit(l) => lits.push(Formula::Atom(l)),
+						BoolViewInner::Lit(l) => {
+							lits.push(Formula::Atom(l));
+							all_true = false;
+						}
 					}
 				}
+				if all_true {
```
Dekker1: Same here.

crates/huub/src/model/bool.rs
AllenZzw: Updated, thanks!

crates/huub/src/model/bool.rs
AllenZzw: Updated.

PR #84, opened 2024-08-16
fix reason for setting lower/upper bounds of the result variable

AllenZzw:
When the domain of index variable contains holes, the propagator for "array_int_var_element" constraint returns the reason containing false literals. This is because some values are not in the original domain of the index variable, and the literal [idx!=i] is constant false. The fix filters out these constant false literals in the reason.

### DIFF
--- a/crates/huub/src/propagator/array_var_int_element.rs
+++ b/crates/huub/src/propagator/array_var_int_element.rs
@@ -38,6 +38,7 @@ impl ArrayVarIntElementBounds {
 	}
 }
 
+// todo: incremental propagation for bounds consistency of the element constraint
 impl<P, E, T> Propagator<P, E, T> for ArrayVarIntElementBounds
 where
 	P: PropagationActions,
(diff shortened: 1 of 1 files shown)


crates/huub/src/propagator/array_var_int_element.rs
```diff
@@ -146,14 +147,19 @@ where
...
-					.map(|(i, &v)| {
+					.filter_map(|(i, &v)| {
 						if actions.check_int_in_domain(self.idx, i as IntVal) {
-							actions.get_int_lit(v, LitMeaning::GreaterEq(min_lb))
+							Some(actions.get_int_lit(v, LitMeaning::GreaterEq(min_lb)))
 						} else {
-							actions.get_int_lit(self.idx, LitMeaning::NotEq(i as IntVal))
+							let l = actions.get_int_lit(self.idx, LitMeaning::NotEq(i as IntVal));
+							if let crate::solver::view::BoolViewInner::Const(false) = l.0 {
+								None
```
Dekker1: Can you explain the reasoning for this change?

It seems you were requesting a literal that is false (and therefore cannot be part of a valid reason), but just ignoring it instead.

PR #92, opened 2024-08-22
fix intersect_int_domain when the input IntView is a constant

AllenZzw:


Dekker1:
Is there a test case for this change? It would be great to check that this is correct when we change the model propagation engine

PR #93, opened 2024-08-22
update constraint for int_eq_reif

AllenZzw:
Update propagator in the solver for int_eq_reif


crates/huub/src/model/flatzinc.rs
```diff
@@ -17,7 +17,7 @@ use pindakaas::{
 };
 use rangelist::{IntervalIterator, RangeList};
 use thiserror::Error;
-use tracing::warn;
+use tracing::{trace, warn};
```
Dekker1: This seems like it is a leftover change.

It seems that the tests have stopped running on your Pull Request again, so it did not reveal these automatically

PR #101, opened 2024-08-27
Fix/conflict lazy literal

AllenZzw:
- return `Err` if `lit_req` in `propagate_int` is conflict with the current domain of an integer variable 
- output warning instead of error for the debug helper to inspect whether a literal is true at the time of propagation
- add simple jobshop test cases from MiniZinc Benchmark


AllenZzw:
Regarding the second point, the assumption that all literals in a reason clause are true is no longer true in the setting lazy literals. When there are multiple domain changes from by a propagator, passing the first domain change literal to SAT solver can make the reason literals for other domain change become false. 

See this [false_reason_log.txt](https://github.com/user-attachments/files/16756709/false_reason_log.txt) for an example. During the last propagation of `disjunctive` constraint, both `X_INTRODUCED_65_≥512` and `X_INTRODUCED_40_≥512` are propagated. When receiving `X_INTRODUCED_65_≥512`, the SAT solver further propagates `X_INTRODUCED_49_<425`, and it makes a reason literal `X_INTRODUCED_49_≥425` for `X_INTRODUCED_40_≥512` becomes false. 

We may consider checking the validity of reason literals somewhere else.

Dekker1:
> Regarding the second point, the assumption that all literals in a reason clause are true is no longer true in the setting lazy literals. When there are multiple domain changes from by a propagator, passing the first domain change literal to SAT solver can make the reason literals for other domain change become false.
> 
> See this [false_reason_log.txt](https://github.com/user-attachments/files/16756709/false_reason_log.txt) for an example. During the last propagation of `disjunctive` constraint, both `X_INTRODUCED_65_≥512` and `X_INTRODUCED_40_≥512` are propagated. When receiving `X_INTRODUCED_65_≥512`, the SAT solver further propagates `X_INTRODUCED_49_<425`, and it makes a reason literal `X_INTRODUCED_49_≥425` for `X_INTRODUCED_40_≥512` becomes false.
> 
> We may consider checking the validity of reason literals somewhere else.

This seems like it is because you are working on the `develop` branch, and not the branch of #71. The changes in that branch will ensure that all clauses defining the lazy literals are defined first, so when you propagate then the reason should again hold, even you are created stronger explanations.

AllenZzw:
> > Regarding the second point, the assumption that all literals in a reason clause are true is no longer true in the setting lazy literals. When there are multiple domain changes from by a propagator, passing the first domain change literal to SAT solver can make the reason literals for other domain change become false.
> > See this [false_reason_log.txt](https://github.com/user-attachments/files/16756709/false_reason_log.txt) for an example. During the last propagation of `disjunctive` constraint, both `X_INTRODUCED_65_≥512` and `X_INTRODUCED_40_≥512` are propagated. When receiving `X_INTRODUCED_65_≥512`, the SAT solver further propagates `X_INTRODUCED_49_<425`, and it makes a reason literal `X_INTRODUCED_49_≥425` for `X_INTRODUCED_40_≥512` becomes false.
> > We may consider checking the validity of reason literals somewhere else.
> 
> This seems like it is because you are working on the `develop` branch, and not the branch of #71. The changes in that branch will ensure that all clauses defining the lazy literals are defined first, so when you propagate then the reason should again hold, even you are created stronger explanations.

I am confused. All changes in #71  has been merged to `develop` branch, what is the difference? 

By "reason literals", I mean the literals in the reason constructed by a propagator instead of literals in the defining clause for a lazy literal. Reading the log, I suspect the problem is that domain may change after the propagation of a lazy literal, which makes some literals in the reason (constructed by a propagator) of another "to-be-propagated" literal become false.

Dekker1:
Sorry I meant #73

Since literals are propagated in order, it should still all hold.

AllenZzw:
I tried to incorporate the changes in #73 and it indeed fixes the issue. 

I wonder whether we can still merge this branch as it is first? Otherwise, the fixes need to be merged with #71, which may make too many changes in just one commit or one pull request.

Dekker1:
Resolved by #71

PR #106, opened 2024-09-02
add new benchmark instances and update benchmark script for fzn-huub

AllenZzw:
- update ``check_final`` function to have different behaviours for optimization and satisfaction instance 
- allow only comparing objective value for optimization instances which may have different optimal solutions in different runs (for example, kidney_exchange instances)
- update sampling mode to `SamplingMode::Flat` for long-running benchmarks (https://bheisler.github.io/criterion.rs/criterion/enum.SamplingMode.html)
- update and add jobshop instances 
- add new instances with a focus on `int_lin_eq`, `int_lin_le` and `array_var_int_element` constraints


Dekker1:
Just had quick look, but a few minor notes:
- It looks like the adjusted version of `jobshop_la04` doesn't use the disjunctive propagator. Is there a reason for this?
- There is an instance `jobshop_simple`, which seems like an instance, we made a while back. It would be better if we use instances with names so it is easy to locate what their origin is, so we can recompile them if necessary. (Maybe we should add comments with the origin of the instances).
- The radiation instances now use dashes instead of underscores, but I would love to just consistently use underscores.
- Regarding the classification: There seem to be a few instances that are shorter in `under_second` than the ones in `instant`. Also a few `under_second` that take more than a second 😅 . I would suggest we add an extra class `a_few_seconds` (or something similar for these.
- Finally, I wonder whether we should consider a even larger class that takes 10-30 seconds to solve.

Thanks for all the hard work! 👍

AllenZzw:
Thanks for the comments! 
- `jobshop_la04` still uses the disjunctive propagator. It seems that the original `jobshop_la04` instance only contains 20 tasks and is not compiled from the `la04` instance which has 50 tasks. It has been renamed into `jobshop_simple`. The `jobshop_la04` is recompiled and added into benchmark. 
- The radiation and svrp instances are renamed to use underscores instead of dashes. 
- An extra class `a_few_seconds` is added. 
- Interestingly, the results from the server is slower than those from the desktop. I have put instances into different groups according to the report from CodSpeed. 
- It is a good idea. I wonder whether we could add this later when the performance is more stale and there are more propagators for global constraints?

Dekker1:
> * `jobshop_la04` still uses the disjunctive propagator. It seems that the original `jobshop_la04` instance only contains 20 tasks and is not compiled from the `la04` instance which has 50 tasks. It has been renamed into `jobshop_simple`. The `jobshop_la04` is recompiled and added into benchmark.

Sorry, I must have looked to quick. I would still prefer to not have a instance that we don't (for sure) know the origin of. Can you find the data file that this used (and change the name), or find a similar instance? 

> * Interestingly, the results from the server is slower than those from the desktop. I have put instances into different groups according to the report from CodSpeed.

I think that is probably expected. The CI machines are likely to be slower.

> * It is a good idea. I wonder whether we could add this later when the performance is more stale and there are more propagators for global constraints?

Okay, let's wait until later when there is more to be benchmarked.

Dekker1:
This is probably also the right time to add a `.gitattributes` file that then lists the `.fzn.json` files as generated:
```
crates/fzn-huub/corpus/*.fzn.json linguist-generated
```

EDIT: I actually included this in #108, so don't worry about this anymore

AllenZzw:
> Can you find the data file that this used (and change the name), or find a similar instance?

This simple instance is simplified from `la04` for testing only. I find a similar instance and replace `jobshop_simple` with `jobshop_newspaper`, which is from https://github.com/hakank/hakank/blob/master/minizinc/jobshop_newspaper.dzn

PR #112, opened 2024-09-04
Feat/incremental element propagator

AllenZzw:
- add new interface `check_int_is_fixed` for `InspectionActions` trait 
- update `array_var_int_element` constraint to enable incremental propagation

### DIFF
--- a/crates/huub/src/propagator/array_var_int_element.rs
+++ b/crates/huub/src/propagator/array_var_int_element.rs
@@ -4,7 +4,7 @@ use crate::{
 	actions::{explanation::ExplanationActions, initialization::InitializationActions},
 	propagator::{conflict::Conflict, int_event::IntEvent, PropagationActions, Propagator},
 	solver::{
-		engine::queue::PriorityLevel,
+		engine::{queue::PriorityLevel, trail::TrailedInt},
 		poster::{BoxedPropagator, Poster, QueuePreferences},
 		value::IntVal,
 		view::IntView,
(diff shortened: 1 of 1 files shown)


crates/huub/src/actions/inspection.rs
```diff
@@ -15,4 +15,5 @@ pub(crate) trait InspectionActions: TrailingActions {
 		}
 	}
 	fn check_int_in_domain(&self, var: IntView, val: IntVal) -> bool;
+	fn check_int_is_fixed(&self, var: IntView) -> bool;
```
Dekker1: This is not necessary. Use `if let Some(_) = actions.get_int_val(...)` instead

crates/huub/src/propagator/array_var_int_element.rs
```diff
@@ -19,20 +19,22 @@ use crate::{
 pub(crate) struct ArrayVarIntElementBounds {
 	vars: Vec<IntView>,                // Variables to be selected
 	y: IntView,                        // The selected variable
```
Dekker1: `y` is a bad name as well. Maybe rename to `result`

crates/huub/src/propagator/array_var_int_element.rs
```diff
@@ -19,20 +19,22 @@ use crate::{
...
-	idx: IntView,                      // Variable that stores the index of the selected variable
+	index: IntView,                    // Variable that stores the index of the selected variable
 	action_list: Vec<(u32, IntEvent)>, // List of x variables that have been modified since the last propagation
```
Dekker1: Action list does not seem to be actually used. Probably just remove it and the usage of it

crates/huub/src/propagator/array_var_int_element.rs
```diff
@@ -19,20 +19,22 @@ use crate::{
...
+	min_support: TrailedInt,           // The minimum support of the selected variable
```
Dekker1: Shouldn't the supports now be checked when you get notified? You only want to wake up if the current support has changed the relevant bound

Dekker1: (or if the result has changed any bound of course)

Dekker1 (changes requested):

crates/huub/src/propagator/array_var_int_element.rs
AllenZzw: updated.

crates/huub/src/propagator/array_var_int_element.rs
AllenZzw: You are right. But some other propagation rules will be triggered in this case.  When a non-support variable `self.vars[i]` is updated, we still want to check whether `i` can be removed from the `self.index` variable. 

If we want to further optimize this by checking whether `i` is in the domain of `self.index` in event notification, we may need the `InspectionActions` instead of the `TrailingActions` to be passed to `notify_event`. Since this requires some underlying architecture changes, I don't include it in this pull request.

AllenZzw: Update the `notify_event` function to filter out unnecessary propagations

Dekker1 (approved):

PR #118, opened 2024-09-10
[WIP] enable generalized explanations for linear propagator 

AllenZzw:
- update `transform_lit` to return the strongest possible `LitMeaning` that is equivalent to the input. 
- add `relaxed_lit` to return the weakest `LitMeaning` that is equivalent to the input. 
- add tests for `transform_lit`, `relaxed_lit` and `rev_transform_lit` functions 
- enable generalized explanation for the `IntLinearLessEqBoundsImpl` propagator


AllenZzw:
There are two functions which we may need to clarify: 
- There are multiple literals that are valid returns for the `transform_lit` function. 
- The `get_int_lit_relaxed` function may return constant true literal in two cases: (1) when the input refers to a newly created lazy literal which has not been assigned true yet, and (2) when the input bounds are out of the original integer domains. The first case may relate to a more general problem on whether the `get_*` interfaces have consistent returns about the state.

Dekker1:
@AllenZzw Do you remember what this status of this is? Did we think this improves the overall performance of the solver?

AllenZzw:
> @AllenZzw Do you remember what this status of this is? Did we think this improves the overall performance of the solver?

This has not been tested on the latest develop branch. In the previous testing, it did not show too much performance improvement. I think we could close this first.

PR #120, opened 2024-09-13
strengthen trailed integer upper bound

AllenZzw:


Dekker1:
Very good point! I had looked at this discrepancy in the past, but never realised that this also influenced the way in which we check whether an integer variable is fixed. It certainly seems like it can influence the runtime!

I'll just need to think a little about the fix. The way eager and lazy integer storage handles the bounds is slightly different. So I'm not sure whether this solves the issue in both cases. We might need to generalise the method slightly and maybe think about trailing the upper bound separately for lazy integers.

This also once again makes me wonder whether the integer variable storage shouldn't be implemented using generics. It would make sense to me to separate the implementation of variables with holes in the domains and ones that don't. Chuffed essentially makes this difference.

Dekker1:
I was working on this (and created an extended version of the changes here: https://github.com/Dekker1/huub/tree/fix/tight_upper_bound). However, both these original changes and my changes seem to have some problems with the `sudoku_p48` instance. On this branch, it triggers an assertion in debug mode. On my branch, multiple solutions are allowed (which is not possible). 

It seems that when tightening the bounds, we need better detection of conflicts as well. I will continue working on this tomorrow

PR #125, opened 2024-09-25
Additional Statistics and SAT Oracle (e.g. Cadical) flags

Dekker1:
This PR adds some additional statistics to the output, namely the number of views extracted from the FlatZinc instance, and the number of variables that is removed through unification.

Additionally, it adds some initial options for the oracle SAT solver that can be set (from the command line).

@AllenZzw Were there any other options that we should add for Cadical that we are already using?


Dekker1:
Just a note that this change also disables restarts by default. As far as I'm aware, restarting is not correct when using non-free search from MiniZinc.
This does, of course, change the performance metrics.

PR #136, opened 2024-11-19
Change the meaning of positive order literals to be `x<i`

Dekker1:
This better matches the default `true` phase/polarity used by various
SAT solvers. The SAT solver searches by setting the earliest created
variable to `true`. However, when this literal has the meaning `x≥i`,
where `i` is close to the lower bound, this is unlikely to have much
effect on the overall problem.

This change in meaning, makes the literal representation more similar to
Chuffed.

### DIFF
--- a/crates/huub/src/solver/engine.rs
+++ b/crates/huub/src/solver/engine.rs
@@ -51,7 +51,7 @@ macro_rules! trace_new_lit {
 			is_eq = matches!($def.meaning, LitMeaning::Eq(_)),
 			val = match $def.meaning {
 				LitMeaning::Eq(val) => val,
-				LitMeaning::GreaterEq(val) => val,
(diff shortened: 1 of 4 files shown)


crates/huub/src/solver/engine.rs
AllenZzw, line 1: Shall we also change `LitMeaning::Less` into `LitMeaning::GreaterEq` in Line 635?

AllenZzw:
Overall LGTM except Line 635 in `engine.rs`. Also, I wonder whether we should also add an option like `--default-phase` for easy experimenting of better phases on different benchmark instances?

Dekker1:
Those are about the integer value of Boolean variables. Changing those around would essentially change `false` to represent `1` and `true` to represent `0`.

PR #137, opened 2024-11-19
Detect all conflicts on `Engine::check_solution`

Dekker1:


Dekker1:
@AllenZzw This was one of the changes that we discussed. It seems to affect the job shop benchmarks mostly negatively, but it's unclear to me whether this is just because of a different search order, and we are unlucky in 2 out of 3 changes. In general, giving more information back to the SAT solver with no change in runtime for `check_solution` would seem good.

Could you run a larger test suite to see whether this is worthwhile or not (independent of propagator deactivation)?

crates/huub/src/solver/engine.rs
```diff
@@ -324,20 +324,22 @@ impl PropagatorExtension for Engine {
...
+		loop {
+			ctx.run_propagators(&mut self.propagators);
+			// No propagation can be triggered (all variables are fixed, so only
+			// conflicts are possible)
+			debug_assert!(ctx.state.propagation_queue.is_empty());
+			// Process propagation results, and accept model if no conflict is
+			// detected
+			if let Some(conflict) = ctx.state.conflict.take() {
+				// Move conflict to clauses before backtrack
+				ctx.state.clauses.push_back(conflict);
```
AllenZzw, line 337: The normal clauses in `state.clauses` and conflict clauses in `state.conflict` seems to be handled differently in `add_external_clause`. Shall we change the `conflict` variable into a vector to ensure the consistent handling of conflicts?

Dekker1: The difference is that `state.clauses` are persistent, and will be kept even when you backtrack. We could change `conflict` to be a vector and capture all conflicts immediately, but it would still require the move to `state.clauses` because we immediately backtrack afterwards in `check_solutions`. (Note that the other difference is that all other clauses and propagations are posted before the conflict, to ensure it will cause conflict in the SAT solver. However, since conflicts are added to the end of the list and there cannot be any propagations, this is not a problem).

So if we stick to my proposed design, where we detect all conflicts in `check_solutions`, but find only the first in `propagate`, then don't think we need to change it.

AllenZzw: I see. Probably the performance change is due to the search heuristic. Let me test it using a larger benchmark and also try the ``--vsids-only`` option.

Dekker1: That would be great! Thanks :)

AllenZzw: Tested with jobshop and openshop instances. The performance seems to remain the same.

Dekker1: If the performance is just the same, then I'm not sure whether the change is worthwhile.

Like we discussed in person: only one of the clauses is used in conflict resolution, so any other given clauses are added to the clause database “as is”. I'm not sure whether this wouldn't be something we would hope to avoid.

I'll at least make another branch that ensures that conflict clauses in `check_solution` are still added as “removable” clauses.

Dekker1:
This only seemed to be effective when using the disabling of propagators. For now I'll close this PR.

PR #151, opened 2025-02-17
Further unification concerns

Dekker1:
- **Ensure subscriptions are moved when unifying variables**
- **Ensure `resolve_alias` results in a `Const` variant when possible**
- **Add integer equality constraint for when unification is impossible**


Dekker1:
Note that I've refrained from implementing a binary `IntEq` propagator. Looking at Chuffed, the propagator is normally still split into two `IntLessEqual` (or `GreaterEqual` in their case). So I decided that it probably is a small optimization that we can do later.

PR #161, opened 2025-03-13
Turn on restart by default

AllenZzw:


Dekker1:
This doesn’t seem correct. Restart would break the normal fixed search. Why does it need to be on by default rather than turning it on with free search or manually?

AllenZzw:
This is to make it consistent with the help message: https://github.com/huub-solver/huub/blob/01ca552b78dbeeb880b1ad4491b3df2b7d19353d/crates/fzn-huub/src/main.rs#L44 
I fine if we decide to change the message instead.

Dekker1:
Good point! I think we had it on by default, but then found it wasn't performing the fixed search correctly. So I changed the default, but must have forgotten to change the message.

PR #162, opened 2025-03-20
add special handling for pseudo-boolean constraints

AllenZzw:
Add special handling for two pseudo-boolean constraints: 
- $b_i + b_j \leq 1 === (\neg b_i) \vee (\neg b_j)$
- $\sum_i b_i \geq 1 === \vee_i b_i$

### DIFF
--- a/crates/huub/src/constraints/int_linear.rs
+++ b/crates/huub/src/constraints/int_linear.rs
@@ -4,7 +4,7 @@
 //! integer decision variables.
 
 use itertools::{Either, Itertools};
-use pindakaas::Lit as RawLit;
+use pindakaas::{ClauseDatabaseTools, Lit as RawLit};
 
 use crate::{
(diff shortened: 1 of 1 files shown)


Dekker1:
I think the idea of special handling is a good idea, but I'm not sure about this implementation. Didn't we talk about using the `BoolLinAggregator` from pindakaas?

I'll see if I can whip up an implementation so you can try and see if it also solves your case.

The main thing would be that it further simplifies the constraints and also allows use to further distinguish the different linear variants if we want to do something with it in the future.

AllenZzw:
Yes, we should definitely try ``BoolLinAggregator``. Just added two most common and useful cases here. We can try to switch to ``BoolLinAggregator`` once we have the interface.

AllenZzw:
Close as this is handled in #163

PR #172, opened 2025-04-16
Add seq_precede_chain and value_precede_chain

kletzi:
Add dedicated implementations for the two symmetry breaking globals, corresponding FlatZinc translations, and test cases.


crates/huub/src/constraints/int_value_precede.rs
```diff
@@ -0,0 +1,1144 @@
...
+impl<S: SimplificationActions> Constraint<S> for IntSeqPrecedeChain {
+	fn simplify(&mut self, actions: &mut S) -> Result<SimplificationStatus, ReformulationError> {
+		let mut ub = 0;
+		for &v in self.vars.iter() {
+			if actions.check_int_in_domain(v, ub + 1) {
+				ub += 1;
+			}
+			actions.set_int_upper_bound(v, ub)?;
+		}
+		//todo this can become more powerful if updated upper bound from previous loop is available
```
Dekker1: I'm not sure whether I understand this TODO. Do you mean what the change in upper bound was, or just that it would be more powerful if this is done multiple times?

crates/huub/src/constraints/int_value_precede.rs
```diff
@@ -0,0 +1,1144 @@
...
+}
+
+impl IntSeqPrecedeChainBounds {
+	/// 3 cases:
+	///
+	/// - Current lower bound is above k - This is the value that required the
+	///   earlier lower bound.
+	/// - k is in the domain of var i - Go one step up and to the next variable.
+	/// - k is not in the domain of var i - i can be anything else, go to next
+	///   variable.
```
Dekker1, line 143: This comment should probably explain what the method does first.

crates/huub/src/constraints/int_value_precede.rs
```diff
@@ -0,0 +1,1144 @@
...
+
+		for v in vars {
+			solver.enqueue_on_int_change(prop, v, IntPropCond::Domain);
+		}
+		solver.enqueue_now(prop);
+	}
+
+	/// Do a full propagation run, requires checking all variables in both
+	/// directions.
+	fn propagate_full<P: PropagationActions>(&mut self, actions: &mut P) -> Result<(), Conflict> {
```
Dekker1: This is only done once, maybe rename it to be `initial_propagation` or something.

crates/huub/src/constraints/int_value_precede.rs
```diff
@@ -0,0 +1,1144 @@
...
+}
+
+impl<P, E> Propagator<P, E> for IntSeqPrecedeChainBounds
+where
+	P: PropagationActions,
+	E: ExplanationActions,
+{
+	#[tracing::instrument(name = "seq_precede_chain", level = "trace", skip(self, actions))]
+	fn propagate(&mut self, actions: &mut P) -> Result<(), Conflict> {
+		if self.initialized {
```
Dekker1: I think I would like it better the other way around. Have a `if !self.initialized { return self.propagate_full(actions); }`, then the rest is more straightforward.

crates/huub/src/constraints/int_value_precede.rs
```diff
@@ -0,0 +1,1144 @@
...
+			.iter()
+			.take(i)
+			.map(|&v| actions.get_int_lit(v, IntLitMeaning::NotEq(self.values[j - 1])))
+			.collect()
+	}
+
+	/// Get the lower bound for the index in values, None if any options outside
+	/// values are still in the domain. Has to exclude values below and above the
+	/// range of values, then all holes, finally values with lower index.
+	fn get_lower_bound<I: InspectionActions>(&self, actions: &mut I, i: usize) -> Option<usize> {
```
Dekker1: The naming of this method is a little confusing with `get_int_lower_bound`. Maybe we can think of something more clear.

crates/huub/src/constraints/int_value_precede.rs
```diff
@@ -0,0 +1,1144 @@
...
+				k -= 1;
+				if actions.get_trailed_int(self.last[k]) < i {
+					return Ok((i as usize, k + 1));
+				}
+			}
+
+			i -= 1;
+			if i < 0 {
+				self.propagate_lower_bound(actions, 0, k)?;
+				return Ok((0, k)); //todo There is a conflict now, but it might need propagation to trigger
```
Dekker1: How hard would it be to explicitly create an `Conflict` to return? Would we have to copy `propagate_lower_bound`?

Dekker1 (commented):
Things are looking good. I've made some small changes to make things a little more consistent with the existing code base (in the hope to make it more maintainable).

What is mostly left to do is to make sure that I understand what is going on and have a look at the “TODO” comments that you've left.

Things that I'm already wondering about (and that we can discuss):
- Many methods have the same names between the two propagators. Are any of them fully the same, and if so which ones?
- Both propagators have an `initialized` flag. Shouldn't we just run the initialization on the model layer and then just use the initialized version during propagation?

In general, it would be nice to have a few more comments in the algorithms to actually explain what they are doing. Even the merged propagators lack comments, so it would be good if we can set a new standard with this one.

kletzi:
I added more comments and renamed some of the functions we talked about.

PR #174, opened 2025-04-29
add options to enable/disable preprocessing and inprocessing techniques in CaDiCal

AllenZzw:


crates/fzn-huub/src/lib.rs
Dekker1, line 1: I generally like the attributes of the struct to be sorted alphabetically or to be grouped semantically (with a remark about the group). That makes it a lot easier to find things when everything grows.

crates/huub/src/reformulate.rs
Dekker1, line 1: Similarly, here we should group or sort the options. I certainly would sort the methods alphabetically

crates/huub/src/lib.rs
```diff
@@ -1104,7 +1104,13 @@ impl Model {
 		let any_slv: &mut dyn Any = slv.oracle.solver_mut();
 		if let Some(r) = any_slv.downcast_mut::<Cadical>() {
 			r.set_option("restart", config.restart() as i32);
+			r.set_limit("preprocessing", config.preprocessing() as i32);
```
Dekker1: Interesting that the limit is set using a `bool`. I would have thought that would be a limit in the sense of the “number of times” something is done or the “amount of time” spent on this.

share/minizinc/solvers/huub.msc
Dekker1, line 1: These flags probably all need `--help` text as well. Maybe that should also mention the solver in which these techniques are supported (and that the flag is otherwise ignored). This is not really necessary atm, but might be a good reminder for the future.

share/minizinc/solvers/huub_expr.msc
Dekker1, line 1: I don't think this file should exist

share/minizinc/solvers/huub.msc
Dekker1: We also probably want to keep the flags here in a sorted order.

Dekker1 (changes requested):
It mostly looks good to me, but there are some small changes before I merge this.

I can make the changes if you don't have time

AllenZzw:
> It mostly looks good to me, but there are some small changes before I merge this.
> 
> I can make the changes if you don't have time

Thanks for the comments! I can make those changes.

PR #180, opened 2025-05-14
Add Not First/Not Last propagators for disjunctive

AllenZzw:
- add not-last propagator for disjunctive
- add detectable precedence propagator for disjunctive
- update the interface for omega-theta tree


crates/huub/src/constraints/disjunctive_strict.rs
```diff
@@ -32,31 +32,81 @@ pub struct DisjunctiveStrict {
...
+enum DisjunctivePropagationRule {
+	/// The edge finding propagation rule.
+	EdgeFinding,
+	/// The not-last propagation rule.
+	NotLast,
+	/// The precedence propagation rule.
+	Precedence,
+}
+
+impl From<u64> for DisjunctivePropagationRule {
```
Dekker1: I would prefer removing these `From` implementations and putting the match statement in the `propagation_rule_from_data` (and the other direction). I think this conversion might otherwise be too easy to trigger (in non-obvious ways.

crates/huub/src/constraints/disjunctive_strict.rs
```diff
@@ -32,31 +32,81 @@ pub struct DisjunctiveStrict {
 	pub(crate) start_times: Vec<IntDecision>,
 	/// Durations of each task.
 	pub(crate) durations: Vec<IntVal>,
+	/// Whether to enable the edge finding propagator.
```
Dekker1: I would point to the specific propagation rule enum members, e.g. `DisjunctiveProagationRule::EdgeFinding`

crates/huub/src/constraints/disjunctive_strict.rs
```diff
@@ -32,31 +32,81 @@ pub struct DisjunctiveStrict {
...
+	/// Defaults to `false`.
+	pub(crate) not_last_prop: Option<bool>,
+	/// Whether to enable the detectable precedence propagator.
+	///
+	/// Defaults to `false`.
+	pub(crate) detectable_precedence_prop: Option<bool>,
+}
+
+#[derive(Clone, Debug, Eq, Hash, PartialEq)]
+enum DisjunctivePropagationRule {
```
Dekker1: Missing documentation for the enum itself. Probably point to the propagator struct where the different algorithms will then also be described.

crates/huub/src/constraints/disjunctive_strict.rs
```diff
@@ -32,31 +32,81 @@ pub struct DisjunctiveStrict {
...
+		}
+	}
 }
 
 #[derive(Debug, Clone, PartialEq, Eq, Hash)]
-/// A propagator for the `disjunctive_strict` constraint using the Strict Edge
-/// Finding algorithm.
-pub struct DisjunctiveStrictEdgeFinding {
+/// A propagator for the `disjunctive_strict` constraint using the Overload Checking,
+/// Edge Finding, Not-First/Not-Last, and Detectable Precedence algorithms.
```
Dekker1: Could you add a few sentences to describe the different algorithms and what they do here? We probably also want to add citations to where they came from.

Note that you can use Markdown in the comments to create different sections.

crates/huub/src/constraints/disjunctive_strict.rs
```diff
@@ -67,22 +117,27 @@ struct OmegaThetaTree {
...
+	/// to the task number.
+	task_no: Vec<usize>,
 }
 
 #[derive(Debug, Clone, PartialEq, Eq, Hash)]
 /// A node structure for the [`OmegaThetaTree`].
 struct OmegaThetaTreeNode {
 	/// Total duration of the tasks under the tree rooted at this node.
-	total_durations: i32,
+	total_durations: i64,
```
Dekker1, line 154: Was there a specific reason why these had to grow to `i64`? Is this just because we accept integers to 64 bits?

crates/huub/src/constraints/disjunctive_strict.rs
```diff
@@ -94,6 +149,53 @@ struct TaskInfo {
...
+	/// Return whether the edge finding propagation will be used when
+	/// creating a [`Solver`] object.
+	pub fn edge_finding_propagation_enabled(&self) -> bool {
+		self.edge_finding_prop.unwrap_or(true)
+	}
+
+	/// Ensure the use of the edge finding propagator when this constraint is
+	/// posted to a [`Solver`] object.
+	///
+	/// Note that this method does not affect whether other propagators
```
Dekker1: For these comments, we should probably use “propagation method” instead of “propagator”, since they are combined in a single struct.

crates/huub/src/constraints/disjunctive_strict.rs
```diff
@@ -94,6 +149,53 @@ struct TaskInfo {
...
+	/// Return whether the not-last propagation will be used when
+	/// creating a [`Solver`] object.
+	pub fn not_last_propagation_enabled(&self) -> bool {
+		self.not_last_prop.unwrap_or(false)
+	}
+
+	/// Ensure the use of the not-last propagator when this constraint is
+	/// posted to a [`Solver`] object.
+	///
+	/// Note that this method does not affect whether other propagators
```
Dekker1: See above

crates/huub/src/constraints/disjunctive_strict.rs
```diff
@@ -94,6 +149,53 @@ struct TaskInfo {
...
+	/// Return whether the detectable precedence propagation will be used
+	/// creating a [`Solver`] object.
+	pub fn detectable_precedence_propagation_enabled(&self) -> bool {
+		self.detectable_precedence_prop.unwrap_or(false)
+	}
+
+	/// Ensure the use of the detectable precedence propagator when this
+	/// constraint is posted to a [`Solver`] object.
+	///
+	/// Note that this method does not affect whether other propagators
```
Dekker1: See above

crates/huub/src/constraints/disjunctive_strict.rs
```diff
@@ -126,140 +228,614 @@ impl<S: SimplificationActions> Constraint<S> for DisjunctiveStrict {
...
+			self.detectable_precedence_propagation_enabled(),
+		);
+
 		Ok(())
 	}
 }
 
-impl DisjunctiveStrictEdgeFinding {
+impl DisjunctiveStrictPropagator {
 	#[inline]
```
Dekker1: All the followin methods are marked "inline", but it is a little unclear why. I think we either need to reason about when we mark them as such, or do some experiments. I think we can remove the inline and the compiler will do the reasoning well enough.

crates/huub/src/constraints/disjunctive_strict.rs
```diff
@@ -126,140 +228,614 @@ impl<S: SimplificationActions> Constraint<S> for DisjunctiveStrict {
...
+	/// Return the propagation rule from the data stored for explanation.
+	fn propagation_rule_from_data(&self, data: u64) -> DisjunctivePropagationRule {
+		(data >> 62).into()
+	}
+
+	/// Propagate overload checking propagation rule
+	fn propagate_overload_checking<P: PropagationActions>(
+		&mut self,
+		actions: &mut P,
+	) -> Result<(), Conflict> {
```
Dekker1: The comments in this method do not really explain the reasoning, just the actions being taken. Can you adjust them so that the method says what "overload checking" means (and how its going to prune)? Then the comments in the method should talk about why things are done. Why do we clear the tree and sort the tasks? What does it mean to check the "overload checking propagation rule"?

crates/huub/src/constraints/disjunctive_strict.rs
```diff
@@ -126,140 +228,614 @@ impl<S: SimplificationActions> Constraint<S> for DisjunctiveStrict {
...
 
-	#[inline]
-	/// Return the (current) latest completion time of task `i`.
-	fn latest_completion_time<I: InspectionActions>(&self, i: usize, actions: &mut I) -> i32 {
-		actions.get_int_upper_bound(self.start_times[i]) as i32 + self.durations[i] as i32
+	/// Propagate detectable precedence propagation rule
+	fn propagate_detectable_precedence<P: PropagationActions>(
+		&mut self,
+		actions: &mut P,
+	) -> Result<bool, Conflict> {
```
Dekker1, line 601: Could you change the comments for this method similar to `propagate_overload_checking`?

crates/huub/src/constraints/disjunctive_strict.rs
```diff
@@ -126,140 +228,614 @@ impl<S: SimplificationActions> Constraint<S> for DisjunctiveStrict {
...
-	fn explain(&mut self, actions: &mut E, _: Option<RawLit>, task_no: u64) -> Conjunction {
-		// explain why the set of tasks Lcut(j) ∪ {i} cannot be completed before lct_j
-		// since energy of the set of tasks (including i) within the time window [earliest_start, latest_completion] is overloaded
-		// explain lower bound propagation for edge finding
-		let task_no = task_no as usize;
+	/// Propagate not-last propagation rule
+	fn propagate_not_last<P: PropagationActions>(
+		&mut self,
+		actions: &mut P,
+	) -> Result<bool, Conflict> {
```
Dekker1, line 864: Could you change the comments for this method similar to `propagate_overload_checking`?

crates/huub/src/constraints/disjunctive_strict.rs
```diff
@@ -126,140 +228,614 @@ impl<S: SimplificationActions> Constraint<S> for DisjunctiveStrict {
...
+			// If for all j in NLset(i) [lst_j ≤ lct_i'], then max{lst_j, j \in \Omega} ≤ lct_i', and lct_i' should be set
+			let (bv, _) = actions
+				.get_int_lit_relaxed(self.start_times[j], IntLitMeaning::Less(updated_lct_i + 1));
+			clause.push(bv);
+		}
+		clause
+	}
+
+	/// Propagate edge finding propagation rule and overload checking
+	fn propagate_edge_finding<P: PropagationActions>(
```
Dekker1: Could you change the comments for this method similar to `propagate_overload_checking`?

crates/huub/src/constraints/disjunctive_strict.rs
```diff
@@ -296,132 +946,67 @@ where
 			})
 			.collect()
 	}
-	#[tracing::instrument(name = "disjunctive_bounds", level = "trace", skip(self, actions))]
+
+	#[tracing::instrument(name = "disjunctive_strict", level = "trace", skip(self, actions))]
 	fn propagate(&mut self, actions: &mut P) -> Result<(), Conflict> {
-		// sort the tasks by earliest start time and construct the EF trees
+		// Sort the tasks by earliest start time and initialize the Omega-Theta tree
```
Dekker1, line 1090: Can you adjust the comment to say why this is done?

crates/huub/src/constraints/disjunctive_strict.rs
```diff
@@ -484,36 +1069,16 @@ impl OmegaThetaTree {
...
-	/// Finding the task responsible for min{est_S, est_i} where
-	/// - S is the set of tasks in the tree
-	/// - task i is one of the gray task in the tree
-	fn gray_est_responsible_task(&self, earliest_completion_time: i32) -> usize {
-		assert!(self.nodes[0].earliest_completion <= earliest_completion_time);
-		assert!(self.nodes[0].earliest_completion_gray >= earliest_completion_time);
+	/// Find the task responsible for pushing the gray task’s earliest completion time (ECT),
+	/// i.e., ECT(Ω ∪ i) > time_bound.
+	fn blocking_task(&self, time_bound: i64) -> usize {
+		assert!(self.nodes[0].earliest_completion <= time_bound);
```
Dekker1: Should this not use `self.root()` instead of `self.nodes[0]`?

crates/huub/src/constraints/disjunctive_strict.rs
```diff
@@ -610,17 +1230,17 @@ impl OmegaThetaTree {
 		let right_earliest_completion_gray = self.nodes[right_child].earliest_completion_gray;
 
 		self.nodes[i].total_durations = left_total_durations + right_total_durations;
-		self.nodes[i].earliest_completion = i32::max(
+		self.nodes[i].earliest_completion = i64::max(
```
Dekker1: We shouldn't use `i64::max` (nor `i32::max`). Instead we should use either the method call `right_earliest_completion.max(left_earliest_completion + right_total_durations)`. Or otherwise, we should import `std::cmp::max` and that `max` function (or `cmp::max` if you want to make it explicit).

crates/huub/src/flatzinc.rs
```diff
@@ -1207,7 +1207,22 @@ where
 							.iter()
 							.map(|l| self.par_int(l))
 							.try_collect()?;
-						self.prb += disjunctive_strict(start_times, durations);
+
+						let force_ef_prop =
```
Dekker1: - `force_ef_prop` -> `force_edge_finding`
- `force_nl_prop` -> `force_not_last`
- `force_dp_prop` -> `force_detectable_precedence`

Dekker1 (changes requested):
The PR looks excellent, but I hope you won't mind taking this opportunity to improve the documentation.

I'm also missing the definitions of the annotations you're detecting in `flatzinc.rs`. From memory, these do not exist in the MiniZinc standard library, so we need to add definitions for them in our solver library. Generally, this means adding a `huub.mzn` file in the solver library with these extended definitions.

crates/huub/src/constraints/disjunctive_strict.rs
AllenZzw: Added description in the corresponding function for each propagation rule.

crates/huub/src/constraints/disjunctive_strict.rs
AllenZzw: yes, because durations are i64 in the modelling layer.

crates/huub/src/constraints/disjunctive_strict.rs
AllenZzw: it is a requirement for the Omega-Theta tree. See its description for more details.

PR #194, opened 2025-07-16
fix(huub): resolution of a sequence of linear transformation aliases

kletzi:
I found a bug when resolving multiple stages of linear transformations after variable unifications. The problem was as follows:

- resolve_alias is called for a linear transformation -x-1
- x is an alias for y+1
- y is an alias for z+1

The result was that the final offset was incorrectly calculated which lead to wrong bounds being returned. I added a test case replicating this chain (which initially failed), and a fix in resolve_alias which makes the test case work.


Dekker1 (approved):
Great catch! Thank you for the fix and the helpful test case!

PR #195, opened 2025-07-16
Sweep based DiffN Propagator

dannem1337:
A Sweep based DiffN Propagator created within [this master thesis](https://urn.kb.se/resolve?urn=urn:nbn:se:uu:diva-562628).

**What's been added**
- DiffN propagator with explanations 
- Integrations with minizinc
- Optimizations: Merge, Overhang and Sources and Targets
- Evaluations comparing with decomposition

**TODOs**
- **BUG:** From what I have seen, whenever a propagation determines the model to be unsatisfiable, it is deemed that we are constructing an empty conflict. Although, looking at the logs, the explanations seem valid. The regression test test_diffn_unsat and bug.mzn should capture this error. The following show the explanations for when this error could occur.
```
   0.005409190s TRACE propagate{level=0}:diffn: huub::constraints::int_diffn: reason of obj 0 in dimension 0 [1..1] >= 1
   0.005414600s TRACE propagate{level=0}:diffn: huub::constraints::int_diffn: reason of obj 0 in dimension 1 [1..1] < 2
   0.005417645s TRACE propagate{level=0}:diffn: huub::constraints::int_diffn: reason of obj 0 in dimension 1 [1..1] >= 1
   0.005420625s TRACE propagate{level=0}:diffn: huub::constraints::int_diffn: reason of obj 1 in dimension 0 [1..1] < 5
   0.005440054s TRACE propagate{level=0}:diffn: huub::constraints::int_diffn: reason of obj 1 in dimension 0 [1..1] >= -2
   0.005457176s TRACE propagate{level=0}:diffn: huub::constraints::int_diffn: reason of obj 1 in dimension 1 [1..1] < 5
   0.005463746s TRACE propagate{level=0}:diffn: huub::constraints::int_diffn: reason of obj 1 in dimension 1 [1..1] >= -2
```
-  The explanation for conflict could probably be improved. Currently conflict is only reasoned about in one dimensions though, it occurs in all dimensions since a conflict only occurs when a forbidden regions fully overlaps the active object.
-  `all_fr` in `generate_fr` could be moved to self to reduce the number of allocations


crates/huub/Cargo.toml
```diff
@@ -19,6 +19,7 @@ pindakaas = { git = "https://github.com/pindakaashq/pindakaas.git", features = [
 	"external-propagation",
 ] }
 rangelist = "0.2"
+smallvec = "1.15.0"
```
Dekker1: Did you measure the performance difference when using `smallvec`. I'd prefer to keep the number of dependencies as small as possible. So I'd prefer to avoid adding a new one unless necessary.

crates/huub/src/flatzinc.rs
```diff
@@ -1122,6 +1122,78 @@ where
...
+							.map(|x| x.iter().map(|l| self.lit_int(l)).collect())
+							.try_collect()?;
+
+						let box_size: Vec<Vec<_>> = dx
+							.iter()
+							.zip(dy.iter())
+							.map(|(a, b)| vec![a, b])
+							.map(|x| x.iter().map(|l| self.lit_int(l)).collect())
+							.try_collect()?;
+						if is_nonstrict {
```
Dekker1: Looks like this `if {} else {}` is not required, and `is_nonstrict` can be used as an argument for `diffn_int`.

crates/huub/src/flatzinc.rs
```diff
@@ -1122,6 +1122,78 @@ where
...
+							.map(|c| c.to_vec())
+							.collect();
+
+						let sizes = self.arg_array(box_size)?;
+						let sizes: Vec<_> = sizes.iter().map(|l| self.lit_int(l)).try_collect()?;
+						let sizes: Vec<Vec<_>> = sizes
+							.chunks(dimensions.try_into().unwrap())
+							.map(|c| c.to_vec())
+							.collect();
+						if is_nonstrict {
```
Dekker1: Replace `if` statement with using `is_nonstrict` as an argument.

share/minizinc/solvers/huub.msc
```diff
@@ -4,7 +4,7 @@
   "id": "solutions.huub",
   "tags": ["int", "bool", "sat", "lcg"],
   "inputType": "JSON",
-  "executable": "../../fzn-huub",
+  "executable": "../../../target/release/fzn-huub",
```
Dekker1: This will conflict with the current `develop` branch. This change should be removed.

crates/huub/src/constraints/int_diffn.rs
```diff
@@ -0,0 +1,829 @@
...
+	solver::{
+		activation_list::IntPropCond, queue::PriorityLevel, trail::TrailedInt, BoolView, IntView,
+	},
+	IntDecision, IntLitMeaning, IntVal,
+};
+
+#[derive(Debug, Clone, PartialEq, Eq, Hash)]
+/// Representation of the `diffn_int` constraint within a model.
+///
+/// This constraint enforces that all k-dimensional rectangles does no overlap
```
Dekker1: Small spelling mistake "do not overlap"

crates/huub/src/constraints/int_diffn.rs
```diff
@@ -0,0 +1,829 @@
...
+	pub(crate) box_posn: Vec<Vec<IntDecision>>,
+	/// The sizes of all objects in all dimensions
+	pub(crate) box_size: Vec<Vec<IntDecision>>,
+	/// True if non_strict diffn is used, i.e ignore all rectangles
+	/// with a size of 0
+	pub(crate) non_strict: bool,
+}
+
+#[derive(Debug, Clone, PartialEq, Eq, Hash)]
+/// Sweep based propagator for the `diffn_int` constraint.
```
Dekker1: It would be great if we can add some additional documentation here with the main idea of what the propagator does and a reference(s) to the paper(s) that describe it.

crates/huub/src/constraints/int_diffn.rs
```diff
@@ -0,0 +1,829 @@
...
+pub struct IntDiffnSweep {
+	/// The origin positions of all objects in all dimensions
+	box_posn: Vec<Vec<IntView>>,
```
Dekker1: I wonder if it would help to use a different structure instead of `Vec<Vec<_>>` for each of these members. The outer vector of all these would be the same size.

It might make sense structurally to have an `DimStore` type that contains all data for that specific dimension, and then store a `Vec` with length `dimension` of those.
With Rust doing dynamic checking of indexes, it might give a small performance boost.

crates/huub/src/constraints/int_diffn.rs
```diff
@@ -0,0 +1,829 @@
...
+	/// Are all sizes fixed
+	fixed_sizes: bool,
+	/// Tracks the upper bound of the positions
+	ub_tracker: Vec<Vec<IntVal>>,
+	/// Tracks the lower bound of the positions
+	lb_tracker: Vec<Vec<IntVal>>,
+	/// Tracks the lower bound of the sizes
+	lb_sizes: Vec<Vec<IntVal>>,
+	/// True if non_strict diffn is used, i.e ignore all rectangles with a size of 0
+	non_strict: bool,
```
Dekker1: This seems to never change once the propagator is created, so it should probably be a const generic. (So we create the type as `IntDiffnSweep<const STRICT: bool>`).

crates/huub/src/constraints/int_diffn.rs
```diff
@@ -0,0 +1,829 @@
...
+					}
+				}
+				if all_fixed {
+					let _ = actions.set_trailed_int(self.target[o_idx], 1);
+				}
+			}
+		}
+
+		// Source optimisations
+		let mut active_b = ForbiddenRegion {
```
Dekker1: It feels this object can be allocated in the propagator struct (with the right size), and then never has to be resized.

crates/huub/src/constraints/int_diffn.rs
```diff
@@ -0,0 +1,829 @@
...
+#[cfg(test)]
+mod tests {
+	use itertools::Itertools;
+	use pindakaas::solver::cadical::PropagatingCadical;
+	use tracing_test::traced_test;
+
+	use crate::{diffn_int, reformulate::InitConfig, Decision, Model};
+	#[test]
+	#[traced_test]
+	fn test_diffn_unsat_1() {
```
Dekker1: It would be great to have one or two more test cases. Maybe one or two SAT test cases where the number of valid solutions would be limited.

crates/huub/src/constraints/int_diffn.rs
```diff
@@ -0,0 +1,829 @@
...
+		);
+
+		for v in box_posn.into_iter().flatten() {
+			solver.enqueue_on_int_change(prop, v, IntPropCond::Bounds);
+		}
+        solver.enqueue_now(prop);
+	}
+
+	/// Prune the lower bounds of the domain
+	fn prune_min<P: PropagationActions>(
```
Dekker1: It would be great if for the (inner) propagation methods (like `prune_min`, `prune_max`, `adjust_sweep_min`, `generate_fr`, etc.) we could add comments that really explain the reasoning, not just the actions being taken. Can you adjust them so that the method says what and how it's going to prune? Then the comments in the method should talk about why things are done and what assumptions are made at different parts of the code.

crates/huub/src/constraints/int_diffn.rs
```diff
@@ -0,0 +1,829 @@
...
+			} else {
+				// Reset sweep-point
+				sweep[rotation] = curr_obj_lb[rotation];
+			}
+		}
+		sweep[curr_dimension] = curr_obj_ub[curr_dimension] + 1;
+		false
+	}
+
+	/// Adjusts the sweep and jump point when pruning the upper bound
```
Dekker1: Do we explain somewhere what the sweep and jump point are and what they are used for?

crates/huub/src/constraints/int_diffn.rs
```diff
@@ -0,0 +1,829 @@
...
+			if curr_obj_lb[d] > fr.ub[d] || curr_obj_ub[d] < fr.lb[d] {
+				return false;
+			}
+		}
+		true
+	}
+
+	/// Generates forbidden regions given object o, forbidden regions that do not overlap with
+	/// the starting domain of o are not included
+	fn generate_fr<P: PropagationActions>(
```
Dekker1: Name might better focus on the "forbidden region" part instead of the "generate". Maybe `gen_forbidden_regions`

crates/huub/src/constraints/int_diffn.rs
```diff
@@ -0,0 +1,829 @@
...
+				|| self.lb_tracker[curr_obj_idx][d] > bounding_box.ub[d]
+			{
+				return true;
+			}
+		}
+		false
+	}
+
+	/// Explains all forbidden regions
+	fn explain_fr<P: PropagationActions>(
```
Dekker1: expand name from `explain_fr` to `explain_forbidden_regions`

crates/huub/src/constraints/int_diffn.rs
```diff
@@ -0,0 +1,829 @@
...
+				}
+				let mut possible_ub = self.ub_tracker[o_idx][d];
+				let origin_ub = self.ub_tracker[curr_obj_idx][d];
+
+				let mut possible_lb = self.lb_tracker[o_idx][d];
+				let origin_lb = self.lb_tracker[curr_obj_idx][d];
+
+				if all_fr[fr].ub[d] > origin_ub {
+					possible_lb =
+						origin_ub - actions.get_int_lower_bound(self.box_size[o_idx][d]) + 1;
```
Dekker1: Should this use the `self.lb_sizes`?

crates/huub/src/constraints/int_diffn.rs
```diff
@@ -0,0 +1,829 @@
...
+	/// Generates forbidden regions given object o, forbidden regions that do not overlap with
+	/// the starting domain of o are not included
+	fn generate_fr<P: PropagationActions>(
+		&self,
+		actions: &mut P,
+		fr_support: &mut Vec<usize>,
+		o_idx: usize,
+		dimensions: usize,
+	) -> Option<(Vec<ForbiddenRegion>, Vec<ForbiddenRegion>)> {
+		// TODO: To avoid uneccesary allocations, this could be moved to self
```
Dekker1: Good idea! 👍

crates/huub/src/constraints/int_diffn.rs
```diff
@@ -0,0 +1,829 @@
...
+			let mut fr_support: Vec<usize> = Vec::new();
+
+			if let Some((all_fr, all_fr_explain)) =
+				self.generate_fr::<P>(actions, &mut fr_support, o_idx, self.dimensions)
+			{
+				if self.fixed_in_all_dimensions(o_idx) {
+                    trace!("CONFLICT");
+					// Conflict will occur here since there exists forbidden regions for a fixed
+					// object
+					// TODO: Conflict occurs in all dimensions but we only reason about it in one
```
Dekker1: You are right. You can point at each dimension as cause of the conflict. However, once you resolve one (allow movement in one dimension) then no more conflict exists.

Potentially, this means we could output all conflict clauses or be smart in some other way, but the present strategy of picking the first dimension seems good to me. We can probably remove the TODO.

crates/huub/src/constraints/int_diffn.rs
```diff
@@ -0,0 +1,829 @@
...
+			active_b.lb.push(i64::MAX);
+			active_b.ub.push(i64::MIN);
+		}
+
+		for o_idx in 0..self.box_posn.len() {
+			if actions.get_trailed_int(self.target[o_idx]) == 1 {
+				continue;
+			}
+			for i in 0..self.dimensions {
+				active_b.lb[i] = cmp::min(active_b.lb[i], self.lb_tracker[o_idx][i]);
```
Dekker1: Does `active_b.lb[i].min(self.lb_tracker[o_idx])` not look nicer?

Dekker1 (changes requested):
As discussed, the propagator is looking good. It would be great to get some more documentation to really make it as self-contained as possible.

Thanks again for all the hard work!

crates/huub/src/constraints/int_diffn.rs
```diff
@@ -0,0 +1,884 @@
...
+	pub(crate) box_size: Vec<Vec<IntDecision>>,
+	/// True if non_strict diffn is used, i.e ignore all rectangles
+	/// with a size of 0
+	pub(crate) non_strict: bool,
+}
+
+#[derive(Debug, Clone, PartialEq, Eq, Hash)]
+/// Sweep based propagator for the `diffn_int` constraint.
+///
+/// This propagator was first proposed in "Sweep as a Generic Pruning Technique
```
CervEdin:
```suggestion
/// This propagator was originally proposed in "Sweep as a Generic Pruning Technique
```

dannem1337:
Sorry for the delay, this should fix everything we discussed during the meeting! :D What has been added:
- Non-strict testcase
- store sizes and origins in one Dimstore struct
- Remove unused case coalesce

Dekker1 (approved):

Dekker1:
Sorry that this took a lot longer than it probably should have. I finally freed up some time to refactor the code base. I've updated it to the latest changes in the solver, and I've made a few changes to the data structures. (I think I put you on the wrong track with the `DimStore`. That was not what I meant). Furthermore, I went through some of the naming and documentation and tried to improve where I could.

@dannem1337, I would appreciate it if you could have another look through the codebase, just to see whether there are any obvious problems that you can find or whether there are any misunderstandings that made it into the documentation.

Afterwards, I think this will be ready to merge.

crates/huub/src/constraints/int_diffn.rs
```diff
@@ -72,122 +80,115 @@ where
...
 		self.propagate(ctx)?;
 
-		if self.boxes.iter().all(|b| {
-			b.origins
-				.iter()
-				.chain(&b.sizes)
-				.all(|v| v.val(ctx).is_some())
-		}) {
+		if self.origin.iter_elem().all(|v| v.val(ctx).is_some())
+			&& self.origin.iter_elem().all(|v| v.val(ctx).is_some())
```
dannem1337: These conditions in the if statement look the same, did you mean 
```self.size.iter_elem().all(|v| v.val(ctx).is_some())```

Though, would it not be sufficient for the first condition to be true to make it susbsumed?  The propagator will never prune the domains of the sizes

File: crates/huub/src/constraints/int_diffn.rs [hunk 21]
```diff
@@ -113,51 +121,167 @@ where
...
+	///
+	/// # Returns
+	///
+	/// - `Some(Ordering::Less)`: `other` is a subset of `self`.
+	/// - `Some(Ordering::Greater)`: `self` is a subset of `other`.
+	/// - `Some(Ordering::Equal)`: The regions are identical.
+	/// - `None`: The regions cannot be coalesced because they are disjoint,
+	///   partially overlapping, or touching in a way that doesn't form a subset
+	///   relationship across all dimensions.
+	fn coalensce(&self, other: &Self) -> Option<Ordering> {
```
dannem1337: typo? Should be `coalesce`

crates/huub/src/constraints/int_diffn.rs
```diff
@@ -50,23 +58,23 @@ pub struct IntDiffnSweep<const STRICT: bool, I> {
 	/// since it will not affect any other rectangle.
 	source: Box<[TrailedInt]>,
```
dannem1337, line 61: Just curious here, what becomes the difference between using a boxed array and a Vector here? Both are allocated on the heap right?

dannem1337:
A bug seems to have been introduced in 683c6a6a3314b8bb59945b3016206bbc9516f6bb that didn't exist before

The following model gives a result where some rectangles are overlapping, my guess is that is that something something is wrong when checking conflict. I will continue to investigate this tomorrow
```mzn
include "diffn.mzn";

int: n = 14;
int: maxLen = 100;
array[1..n] of 1..n: Squares = [i | i in 1..n];

var n..maxLen: w;
var n..maxLen: h;
var int: rec = w*h;
array [1..n] of var 0..maxLen: X;
array [1..n] of var 0..maxLen: Y;
constraint forall(i in 1..n) (X[i] <= w-Squares[i] /\ Y[i] <= h-Squares[i]);
constraint diffn(X, Y, Squares, Squares);
solve minimize rec;
output ["w:\(w) h:\(h) X:\(X) Y:\(Y) rec: \(rec) squares: \(Squares)"];
```
Here's a picture of what it generates
coordinates

Dekker1:
Great catch! I think I found the problem. It appears that I thought that I rewrote `Region::coalensce` to return similar to `.cmp`, but it was the other way around. I've pushed a fix, which for me works on the instance you posted. I checked using the following extra output:
```
output
[ let {array[int] of bool: overlap = [fix(X[i]) + Squares[i] <= fix(X[j]), fix(Y[i]) + Squares[i] <= fix(Y[j]), fix(X[j]) + Squares[j] <= fix(X[i]), fix(Y[j]) + Squares[j] <= fix(Y[i])] } in if not(exists(overlap)) then "\(i) overlaps with \(j)\n" else "" endif | i,j in 1..n  where i < j];
```
That initially gave me the overlap, and after the fix no longer does. (I didn't have the fancy visualizations).

crates/huub/src/constraints/int_diffn.rs
```diff
@@ -0,0 +1,986 @@
...
+	I2: ModelIntView<E>,
+{
+	fn simplify(
+		&mut self,
+		ctx: &mut E::PropagationCtx<'_>,
+	) -> Result<SimplificationStatus, E::Conflict> {
+		self.propagate(ctx)?;
+
+		if self.origin.iter_elem().all(|v| v.val(ctx).is_some())
+			&& self.origin.iter_elem().all(|v| v.val(ctx).is_some())
```
CervEdin: This looks like a typo/bug. I'm guessing it's not supposed to check `origin` twice?

```suggestion
		if self.origin.iter_elem().all(|v| v.val(ctx).is_some())
			&& self.size.iter_elem().all(|v| v.val(ctx).is_some())
```

crates/huub/src/constraints/int_diffn.rs
```diff
@@ -0,0 +1,986 @@
...
+			let forbidden_regions = self.forbidden_regions(ctx, obj);
+			if !forbidden_regions.is_empty() {
+				// Check for conflicts: if an object is fixed and is in a forbidden
+				// region, it's a conflict.
+				if self.fixed_object(obj) {
+					let reason = self.explain_propagation(ctx, &forbidden_regions, obj, 0, false);
+					// Trigger a conflict by increasing the lower bound.
+					self.origin[[obj, 0]].set_lower_bound(
+						ctx,
+						self.origin_lb[[obj, 0]] + 1,
```
CervEdin, line 690: apropos the TODO is description
> The explanation for conflict could probably be improved. Currently conflict is only reasoned about in one dimensions though, it occurs in all dimensions since a conflict only occurs when a forbidden regions fully overlaps the active object.

I asked Claude Code to review the code and it has the following comments about this

>  The current explanation using `explain_propagation(ctx, &forbidden_regions, obj, 0, false)` only includes the upper bound for dimension 0 (since it's about to prune the lower bound). But for a conflict, we want to show that the entire fixed position (both bounds in all dimensions) is infeasible.
>
>  This leads to stronger learned clauses because the SAT solver learns "these specific bounds in ALL dimensions together are infeasible" rather than just learning about dimension 0.

I don't know the code/algorithm well enough to make an absolute judgement of the above (it may be totally incorrect) but it looks like it could be legit so I figured I might as well share it.

And it's suggesting to do something like this instead

```rust
  if self.fixed_object(obj) {
      // Build a complete explanation for all dimensions
      let mut reason = Vec::new();

      for d in 0..self.num_dimensions() {
          if TypeId::of::<I2>() != TypeId::of::<IntVal>() {
              reason.push(self.size[[obj, d]].lower_bound_lit(ctx));
          }
          // Include both bounds for all dimensions to show complete fixation
          reason.push(
              self.origin[[obj, d]]
                  .lit(ctx, IntLitMeaning::GreaterEq(self.origin_lb[[obj, d]])),
          );
          reason.push(
              self.origin[[obj, d]]
                  .lit(ctx, IntLitMeaning::Less(self.origin_ub[[obj, d]] + 1)),
          );
      }
      reason.extend(self.explain_forbidden_regions(ctx, &forbidden_regions, obj));

      self.origin[[obj, 0]].set_lower_bound(
          ctx,
          self.origin_lb[[obj, 0]] + 1,
          reason,
      )?;
  }
```

CervEdin: actually it's suggesting to do this, create a new method

```
  Option 1: Dedicated conflict explanation method

  Add a new method that explains conflicts more completely:

  fn explain_conflict<Ctx>(
      &mut self,
      ctx: &mut Ctx,
      forbidden_regions: &[(Region, usize)],
      obj: usize,
  ) -> Vec<Ctx::Atom>
  where
      Ctx: ReasoningContext + ?Sized,
      I1: IntDecisionActions<Ctx>,
      I2: IntInspectionActions<Ctx>,
  {
      let mut reason = Vec::new();

      // Include BOTH bounds for ALL dimensions since the object is fixed
      for d in 0..self.num_dimensions() {
          // Size bounds if needed
          if TypeId::of::<I2>() != TypeId::of::<IntVal>() {
              reason.push(self.size[[obj, d]].lower_bound_lit(ctx));
          }

          // Both lower and upper bounds showing the object is fixed
          reason.push(
              self.origin[[obj, d]]
                  .lit(ctx, IntLitMeaning::GreaterEq(self.origin_lb[[obj, d]])),
          );
          reason.push(
              self.origin[[obj, d]]
                  .lit(ctx, IntLitMeaning::Less(self.origin_ub[[obj, d]] + 1)),
          );
      }

      // Explain the forbidden regions covering this position
      reason.extend(self.explain_forbidden_regions(ctx, forbidden_regions, obj));
      reason
  }

  Then use it:

  if self.fixed_object(obj) {
      let reason = self.explain_conflict(ctx, &forbidden_regions, obj);
      // Trigger conflict
      self.origin[[obj, 0]].set_lower_bound(
          ctx,
          self.origin_lb[[obj, 0]] + 1,
          reason,
      )?;
  }
```

CervEdin (commented):
- a possible bug identified
- a possible fix of TODO proposed

also, fixed a typo here:
https://github.com/dannem1337/huub/pull/3

crates/huub/src/constraints/int_diffn.rs
dannem1337: Yes! Just thinking if we even have to check size? Now when I have thought about it a bit I am leaning towards that we might want check size before reporting SUBSUMED. We wont prune domains of the size variable but we might report conflicts because of it.  So unless all sizes are fixed, we can't be 100% sure we have a valid diffn.

crates/huub/src/constraints/int_diffn.rs
dannem1337: Yes, that is something I have been thinking of! But I think my what I meant with that in these places
```rs
      self.origin[[obj, 0]].set_lower_bound( // <- HERE
          ctx,
          self.origin_lb[[obj, 0]] + 1, // <-HERE
          reason,
      )?;
```
We always apply the conflict to dimension 0, but when we have a conflict it would be equally correct to apply it to dimension 1 like this  
```rs
      self.origin[[obj, 1]].set_lower_bound( // <- HERE
          ctx,
          self.origin_lb[[obj, 1]] + 1, // <-HERE
          reason,
      )?;
```
Or dimension 2 if we have that. My thought was that if we could do some trick with the clauses to try to combine these to something better.

crates/huub/src/constraints/int_diffn.rs
CervEdin: I didn't think about it a lot, besides noting the code is checking the same thing twice.
Maybe it should just check `self.origin` once?

crates/huub/src/constraints/int_diffn.rs
CervEdin: Key Differences

  Memory Layout:
  - `Vec<T>` stores: pointer + length + capacity (3 words = 24 bytes on 64-bit)
  - `Box<[T]>` stores: pointer + length (2 words = 16 bytes on 64-bit)

  Flexibility:
  - `Vec<T>` can grow/shrink (push, pop, resize, etc.)
  - `Box<[T]>` has a fixed size after creation - no capacity for growth

  When to use each:
  - `Vec<T>`: When you need dynamic sizing
  - `Box<[T]>`: When the size is known upfront and won't change

CervEdin: but yes, both are allocated on heap

dannem1337: I see, thanks!

crates/huub/src/constraints/int_diffn.rs
Dekker1: Here, we actually check whether the propagator can be removed. Even if we don't propagate the sizes themselves, we still have to keep the propagator when they are not fixed, as setting them to the wrong value might make the problem unsatisfiable.

I think there might be something smarter that we can do (as you did with the target/source optimization), but checking whether both the origins and sizes are fixed, is at least correct.

crates/huub/src/constraints/int_diffn.rs
Dekker1: Since these clauses would both result in a conflict, it doesn't really matter which one you pick. You could add both, but the solver is currently not really setup to deal with multiple conflict clauses (and would just pick the first one, as CDCL only uses one). It feels like more general future work to investigate whether adding multiple clauses causing a conflict could help.

We did a brief investigation of this when we tried disabling propagators, but the results were very mixed.

For now, I think we can merge the version as is.

PR #200, opened 2025-07-24
Update testing and benchmarking instances

AllenZzw:
- add instances with cumulative global constraints 
- remove `jobshop_la01` due to the system call issue (e.g. [report](https://codspeed.io/huub-solver/huub/branches/Dekker1%3Afeat%2Frustc_hash))


Dekker1:
The times for the benchmarks seems fine, but the `.fzn.json` files are much bigger than any of the other files in our corpus. Each one is at least a megabyte, and some are almost twenty. @AllenZzw What is the size of the files when we do have the `cumulative` global? Keeping the git history in mind, it might be worth only merging after we merge the propagator.

AllenZzw:
> The times for the benchmarks seems fine, but the `.fzn.json` files are much bigger than any of the other files in our corpus. Each one is at least a megabyte, and some are almost twenty. @AllenZzw What is the size of the files when we do have the `cumulative` global? Keeping the git history in mind, it might be worth only merging after we merge the propagator.

All are under 100K if we have the `cumulative` global
```
mrcpsp_j10_10_1.fzn.json 61K 
mrcpsp_j10_10_2.fzn.json 53K 
mrcpsp_j10_10_3.fzn.json 43K 
rcpsp_01.fzn.json 89K 
rcpsp_02.fzn.json 20K 
rcpsp_03.fzn.json 21K 
```

Dekker1:
That's much better. I'll keep this PR open while we're working on the `cumulative` propagator (so we can remember the codspeed timings), but let's not merge the non-cumulative `fzn.json` files into develop.

AllenZzw:
Close this as the instances will be introduced in #188

PR #210, opened 2025-08-12
feat(huub): allow unbounded variable definitions in FlatZinc

Dekker1:


Dekker1:
This resolved #208.

@dannem1337 Sorry I didn't accept your PR. I wanted to add some tests, output a warning, and resolve the issue that the full domain couldn't be `IntVal::MIN..=IntVal::MAX` and add a test case. I was going to add it to your PR, but it ended up being quite the adventure with underlying bugs, so I ended up needing a clean branch to tests where the differences came from.

PR #219, opened 2025-09-02
Advisors on literals that can be assigned immediately

kletzi:
I noted that sometimes I would get an error when setting an advisor for a boolean in difference logic for rcpsp_max:
[advisor_panic.txt](https://github.com/user-attachments/files/22091564/advisor_panic.txt)

I realized it was triggered when the literal to observe can immediately be assigned during the add_observed_var call. 
This would call assign_lit in trail.rs before grow_to_boolvar and lead to the error. I moved the add_observed_var call to the end of advise_on_lit_change, in this case the advisor get set up correctly and gets immediately called even if the literal is directly assigned.

I have test cases using difference logic for this, not sure how to extract the problem into a small stand-alone test case though.

### DIFF
--- a/crates/huub/src/constraints/int_all_different.rs
+++ b/crates/huub/src/constraints/int_all_different.rs
@@ -15,7 +15,7 @@ use crate::{
 	solver::{
 		activation_list::{IntEvent, IntPropCond},
 		queue::PriorityLevel,
-		IntLitMeaning, IntView, IntViewInner,
+		IntLitMeaning, IntView,
 	},
 	IntDecision, IntVal,
 };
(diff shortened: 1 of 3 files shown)


Dekker1:
This seems to be the correct fix. However, I think this behaviour from the SAT solver might lead to unexpected behaviour in the future. (Or at least our immediate processing of it).

The problem is that, at this point, not all propagators/advisors would be attached to the notifications of the literal. (In fact, we know that it is only a single one that is, and there might be countless more). All these other would now not be notified.

Dekker1:
Actually, maybe we can do the same. If we find that a decision variable is already fixed, then we might be able to notify the propagator/advisor immediately.

I'll have to think about how the implementation would work, but I think that would be possible.

zayenz:
> Actually, maybe we can do the same. If we find that a decision variable is already fixed, then we might be able to notify the propagator/advisor immediately.
> 
> I'll have to think about how the implementation would work, but I think that would be possible.

As a point of reference, Gecode schedules propagators when they are subscribed depending on the propagation condition and the domain of the variable. See 23.6 "Scheduling when posting" in https://www.gecode.dev/doc-latest/MPG.pdf. For example, when a variable is assigned when subscribed to, the propagator is always scheduled. If the propagation condition is VAL, then the variable needs to be assigned for the scheudling to happen. If the propagation condition is bound or domain, then scheudling happens always.

Propagators can opt out of the initial scheduling if they want.

In comparison to propagators, an Advisor is never schedule when it is subscribed (27.1 Advisors, third bullet-point), since their invariant is that they are called  upon a domain change to handle that change.

Dekker1:
@zayenz Thank you for the pointers. It does convince me that that is the right way to do it. However, because the current implementation sometimes employs more powerful propagation (or other analysis methods), it might be hard for propagators to find the right spot to check these invariants. Hopefully, I can get things to this point in the near future.

PR #220, opened 2025-09-03
fix(huub): advise on backtrack from artificial decision level

Dekker1:


Dekker1:
@kletzi Can you have a look and see whether this resolves your problem where your propagator wouldn't be notified of backtracking from the “artificial” decision level?

kletzi:
I can confirm that this resolves my problem.

PR #228, opened 2025-09-21
Asking reason lazily in the oracle solver

AllenZzw:
- add option for asking reason eagerly in the oracle solver 
- set the oracle solver to ask reason lazily by default


Dekker1:
The change seems quite straightforward. I would argue to changing to `--reason-eager`.

Did we already run a larger benchmark on this to see its impact?

AllenZzw:
> The change seems quite straightforward. I would argue to changing to `--reason-eager`.
> 
> Did we already run a larger benchmark on this to see its impact?

Tested on instances from the MiniZinc Challenge 2024, with overall positive results.

Naming updated.

PR #235, opened 2025-09-29
fix(huub): trailing in the `BoolBrancher`

Dekker1:
This resolves a problem in the trailing strategy of `BoolBrancher` that
would skip decisions when backtracking.

### DIFF
--- a/crates/huub-cli/corpus/bool_indomain_max.fzn.json
+++ b/crates/huub-cli/corpus/bool_indomain_max.fzn.json
@@ -1,15 +1,29 @@
 {
   "variables": {
-    "x" : { "type" : "bool" },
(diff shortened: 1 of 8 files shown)


Dekker1:
@AllenZzw Once you add a test case as discussed, we can merge this back into develop.

You should be able to push to this branch, but otherwise let me know and I can add the commit to the PR from another place

AllenZzw:
> @AllenZzw Once you add a test case as discussed, we can merge this back into develop.
> 
> You should be able to push to this branch, but otherwise let me know and I can add the commit to the PR from another place

I have updated the testing instances for the Boolean search heuristics. They now include more Boolean variables, which should better cover cases involving skip decisions during backtracking.

PR #238, opened 2025-09-30
fix getting relaxed 'NotEq' literal

AllenZzw:
- fix getting relaxed literal with `NotEq` meaning


Dekker1:
I think we can use stronger literals instead of just the bounds when relaxing the `NotEq` literals. @AllenZzw Can you have a look to see whether the change I pushed still works in your problem case?

AllenZzw:
> I think we can use stronger literals instead of just the bounds when relaxing the `NotEq` literals. @AllenZzw Can you have a look to see whether the change I pushed still works in your problem case?

Pass the test for my instance with alldifferent constraint. Looks good to me.

PR #244, opened 2025-10-23
Example project using Huub as a library

AllenZzw:
Add an example project using Huub as a library for solving jobshop scheduling problem


Dekker1:
I've refactored this some, and it seems in good shape to me now. The main thing that appears to be missing is a README file to tell people about the example and the format that it accepts (and maybe where to find some instance files). It would also be nice to have one or two created instances that we can keep in the repository and run as part of the test suite.

@AllenZzw Could you see whether you could complete this?

AllenZzw:
> I've refactored this some, and it seems in good shape to me now. The main thing that appears to be missing is a README file to tell people about the example and the format that it accepts (and maybe where to find some instance files). It would also be nice to have one or two created instances that we can keep in the repository and run as part of the test suite.
> 
> @AllenZzw Could you see whether you could complete this?

I have updated the example with a README file and some example testing instances. @Dekker1 would be great if you could have a look before merging this.

PR #246, opened 2025-11-03
Reusing `Solver` Propagators for the `Model` layer

### LINKED ISSUE #222: Re-evaluate enqueue when registering subscription/advisor
Considering the pointer from @zayenz

> > Actually, maybe we can do the same. If we find that a decision variable is already fixed, then we might be able to notify the propagator/advisor immediately.
> > I'll have to think about how the implementation would work, but I think that would be possible.
> 
> As a point of reference, Gecode schedules propagators when they are subscribed depending on the propagation condition and the domain of the variable. See 23.6 "Scheduling when posting" in https://www.gecode.dev/doc-latest/MPG.pdf. For example, when a variable is assigned when subscribed to, the propagator is always scheduled. If the propagation condition is VAL, then the variable needs to be assigned for the scheudling to happen. If the propagation condition is bound or domain, then scheudling happens always.
> 
> Propagators can opt out of the initial scheduling if they want.
> 
> In comparison to propagators, an Advisor is never schedule when it is subscribed (27.1 Advisors, third bullet-point), since their invariant is that they are called upon a domain change to handle that change.

The main hold-back for this is the consistency between propagation and analysis in the `Model` and `Solver` layers. For example, it would be unexpected for a `IntEvent::Fixed` advisor to never trigger if it was never fixed in the `Model` because it is immediately fixed when it is created in the `Solver`.

Dekker1:
This PR aims to bring a more coherency and reusability between `Propagator` and `Constraint` in the library code. Their main difference is that `Constraint` is allowed to rewrite itself into other implementations, and it can unify different variables.

In this PR, we make `Propagator` a requirement for `Constraint`. Furthermore, we ensure that `Constraint::simplify` can call `Propagator::propagate`. We also move the code that subscribes to different variables into `Propagator::post`. This avoids that duplication of code.

As a consequence, numerous other things have changed:
- We revisit the generics of `Propagator` and `Constraint`. These now take a (more formal) `ReasoningEngine` generic, that has a list of associated "Contexts" in which different operations are performed.
- Int/Bool variables used by the propagators have become generic. To list the operations required, the different types of actions are now implemented on the different types and available in different context, e.g. `IntVarRef: IntInspectionActions<Ctx>` and `IntView: IntPropagationActions<Ctx>`. I've added the shorthands `SolverIntView<E>` and `ModelIntView<E>` to quickly require most normal operations exposed for `Propagator` and `Constraint` respectively.
- As the `Constraint` implementation now reuse some of the trailing infrastructure, their creation requires the `Model` to set this up. I've changed all "constraint functions" to take `Model` as their first argument. They now no longer require the `+=` to the `Model` afterwards. This is in line with "posting functions" in solvers like Gecode. If the constraint allows setting of options, then a mutable reference to it is exposed to change these options.
- To get `ReasonBuilder` to be generic with different `ReasonEngine` implementations, it is no longer possible to pass a literal directly as the reason. Instead of a `lit` as a reason `[lit]` should work, though.

Questions / TO DO:
- [ ] Make things have changed names or have (slightly) changed semantics. Now might be a good time to think about it:
  - [ ] Creating a new propagator for use in an engine (currently `<PROP>::new`)
  - [ ] Creating and posting a new propagator, and optionally clauses, to enforce a constraint (currently `<PROP>::new_in`)
  - [ ] Trait for context in which propagators can be applied (currently `ReasoningEngine`)
  - [ ] `Model`?
  - [ ] `Solver` / `Engine` / `State` / `SolvingContext`?
  - [ ] Method in `Propagator` called to subscribe it to different decision variables (currently `Propagator::post`)
  - [ ] `PostingContext` / `ModelPostingContext`
  - [ ] `ModelIntView` / `SolverIntView`
  - [ ] `BoolDecision` / `BoolView` / `IntDecision` / `IntView`?
  - [ ] `Constraint::simplify`?
- [x] Finish the `ModelPostingContext` and `Model` queueing infrastructure. Currently, advisors are missing, and setting enqueue to false is not handled.
- [x] Finish the conflict creation methods. Currently, lazy explanations are not handled.
- [x] Add all the missing documentation
- [x] Solve any testing problems (most already succeed, but a few flatzinc tests fail)
- [x] Solve other clippy issues, such as the exposing of `Engine`.
- [ ] Add additional subscription functionality so that `Constraint` implementations can subscribe to when variables are unified.

Other small things that I might include:
- [x] In `cumulative`, I found that the `actions` parameter was often not in the first place. That seems strange, and now might be a good time to fix it.
- [x] In value precede variants, there are recursive functions that seem easy to turn into loops. I might give that a go.

**Next Steps**: Once this PR is merged, the next goal would be to see whether it is worthwhile to perform static dispatch on `IntView` when posting the `Propagator` implementations. This might avoid a lot of runtime dispatching during propagation.

@AllenZzw It would be great if you can have a look and see whether you see any problems with the changes included. Please let me know whether you have any concerns, whether there are things you think need more discussion or thought.

@kletzi Could you let me know whether this encompasses the "intermediate" interface that you had been using? Do you see anything that raises concern or that could be improved?


Dekker1:
@kletzi I would appreciate it if you could have a look at 56970805564ff24fb5c143f6f6e8b5360b1b1847 in particular to ensure that I didn't make any mistakes turning your recursive functions into loops.

Dekker1:
This PR has become a bit of a beast, so I think I'll merge and leave the two todo's as issues.

PR #262, opened 2026-01-12
fix getting lt Boolean view

AllenZzw:
- fix getting `lt` Boolean view 
- add tracing logs in `int_div` simplification and propagation


Dekker1:
Thank you for the fix. I think the `trace!` statements are not yet correct. Let's separate that for a later PR. I'll merge the fix.

PR #269, opened 2026-01-17
fix Boolean formula simplification

AllenZzw:
This issue causes some speck optimization instances in the MiniZinc Challenge 2023 to become unsatisfiable or to produce incorrect objective values.

### DIFF
--- a/crates/huub/src/reformulate.rs
+++ b/crates/huub/src/reformulate.rs
@@ -203,54 +203,80 @@ where
 			};
 			Ok(bv)
 		};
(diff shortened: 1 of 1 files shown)


Dekker1:
Good catch! Any chance there is a unit test that you can think of that triggers this?

AllenZzw:
The `simplify` function needs `PropagationCtx` as an argument. Is there any easy way to create a "mock context" for testing the functions of each propagator?

Dekker1:
`Model` is the `PropagationCtx` for `Model`, so that should work.

PR #270, opened 2026-01-17
update the default domain for unbounded integer variables

AllenZzw:
This issue causes Work Task Variation instances in the MiniZinc Challenge 2025 to become unsatisfiable. The fix updates the default domain for unbounded integer variables to [i32::MIN,i32::MAX] to avoid integer overflow/underflow during `IntLinear` constraint propagation.

### DIFF
--- a/crates/huub/src/flatzinc.rs
+++ b/crates/huub/src/flatzinc.rs
@@ -37,7 +37,7 @@ use crate::{
 
 /// Domain assumed for integer decision variables that do not have a domain
 /// definition.
-const FULL_INT_DOMAIN: RangeInclusive<IntVal> = IntVal::MIN..=IntVal::MAX;
+const FULL_INT_DOMAIN: RangeInclusive<IntVal> = (i32::MIN as i64)..=(i32::MAX as i64);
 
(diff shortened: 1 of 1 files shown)


Dekker1:
This doesn't seem like a correct fix. Where do you find the overflow? That would probably be the right place to fix it. We can discuss this in more detail in person if you like

AllenZzw:
When testing with `generated-seed-3-length-10-open-8-workers-12-block-15.dzn`, the error `attempt to multiply with overflow` is triggered while executing the code at  
https://github.com/huub-solver/huub/blob/ddeda7be8095900d3da28c1beaa63d25b462c48e/crates/huub/src/lib.rs#L1278

Tracing the issue back suggests that an unbounded variable is being used in a `LinearView` struct.

This fix restricts the default domain of unbounded variables to a smaller range to reduce the likelihood of overflow.

Dekker1:
The better solution would instead be to restrict the domain of the `LinearView` to be `IntVal::MIN..=IntVal::MAX` when we create them, which in turn will prune the domain of the underlying variable. That would automatically avoid the overflow.

PR #273, opened 2026-01-27
feat: overflow handling

Dekker1:
- **fix: use overflow aware methods when creating a Model from FlatZinc**
- **style: reorder rust-items in module files**
- **fix: handle overflow in IntPowBounds**
- **feat: handle overflow in IntTimesBounds**
- **feat: handle overflow in integer linear propagation**


Dekker1:
@AllenZzw This should be mostly ready. It resolves the problem with the instance you sent me, but also has a more general approach to integer overflows. I would appreciate it if you could have a look and see whether there is anything that I've overlooked.

PR #284, opened 2026-03-25
fix: out of bounds array access in value precede chain

Dekker1:


Dekker1:
We ran into some issue where the value precede chain propagator would access out-of-bound elements on the vectors. @kletzi could you have a look at the changes and see if they make sense?

kletzi:
I noticed that in the new test case, the solution `x0=0` was missing. More generally, the propagations causing the array access violations should have never been triggered in the first place since they were incorrectly trying to enforce lower bounds in the initial propagation when all variables could take values outside those enforced by the chain.

Therefore, I think the array access should stay as it is (violation indicates wrong propagation), but I added a check for triggering the lower bound propagation and additional test cases to verify correct behaviour.

Currently those changes are pushed on my [own branch](https://github.com/kletzi/huub/tree/fix/value_precede).

Dekker1:
Thank you for having a look. It sounds like my spot fixes really didn't work. I will have a look at your fix instead tomorrow

PR #285, opened 2026-03-27
Early return in cumulative propagator 

AllenZzw:
Early return without propagating resource-usage variables when the domains of start-time variables are updated. This ensures that the time profile is not obsolete when propagating the resource-usage variables.


Dekker1:
I'm glad you found the problem! 😁 

Is it possible to add a regression test for this?

PR #297, opened 2026-03-31
Negation issue on creation of solver lits

kletzi:
I found an issue with an incorrect "optimal" solution for 2017/routing-flexible/routing_GCM_0005.mzn.

I found that the issue was caused by the following pattern: A model `LinearBoolView` of a negated lit `-l` would be mapped to a new solver lit `a` (`a` is equal to the negation of `l`). Then a boolean in the model would look up the mapping of lit `l`, but assume that the positive literal is mapped to `a`. Now different views assume that `a` points to `-l` or `l`.

I added a test case and tried to fix the issue by swapping the mapping of boolean and integer variables. In my understanding, the booleans are always directly mapped positive to positive. Then the later views correctly look up the solver lit and do the transformations. This fix works correctly with the test case and the affected benchmark instance, but please review if this correctly fixes the issue.

### DIFF
--- a/crates/huub/src/lower.rs
+++ b/crates/huub/src/lower.rs
@@ -635,7 +635,7 @@ impl LoweringMapBuilder {
 					None => slv.new_bool_decision().into(),
 				};
 				self.bool_map[idx] = Some(view);
-				view
+				if lit.is_negated() { !view } else { view }
 			}
 			Const(b) => b.into(),
 			IntEq(idx, val) => {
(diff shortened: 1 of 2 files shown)


Dekker1:
Thanks! The problem was indeed that there was a missing negation (and somehow it was done if we had seen a literal with the same variable before). I just fixed it so it always performs the negation correctly.

The idea of creating the integer variables first is actually a conscious decision to try and pack all the ranges of variables required for the mapping in the solver together. Currently, this assumption doesn't matter too much, but I had some ideas of optimizations using this in the future.

PR #354, opened 2026-06-14
feat(circuit): circuit and subcircuit global constraints

AllenZzw:
Add the `circuit` and `subcircuit` global constraints. Three propagators are implemented:
- `no_cycle`: rejects a closed sub-tour that does not span every node, or forbids the closing edge of each maximal fixed chain;
- `scc`: one iterative DFS from the first unfixed root

The implied `alldifferent` is posted as a model-level `IntUnique`.


crates/huub/src/constraints/circuit/graph.rs
```diff
@@ -0,0 +1,142 @@
...
+/// successor variable per node position `0..n`, plus the value `offset` and the
+/// `subcircuit` flag.
+#[derive(Clone, Debug, Eq, Hash, PartialEq)]
+pub(crate) struct CircuitGraph<I> {
+	/// Successor variables, indexed by node position `0..n`.
+	pub(crate) vars: Vec<I>,
+	/// Successor value of node `0`.
+	pub(crate) offset: IntVal,
+	/// Whether this is the `subcircuit` variant.
+	pub(crate) subcircuit: bool,
```
Dekker1: It feels like this could be a generic argument, then all checking of this boolean become free.

crates/huub/src/constraints/circuit/graph.rs
```diff
@@ -0,0 +1,142 @@
...
+	}
+
+	/// Fill `out` with the successor position of every fixed node (`None` for
+	/// unfixed), reusing `out`'s allocation.
+	pub(crate) fn fixed_successors<C>(&self, ctx: &C, out: &mut Vec<Option<usize>>)
+	where
+		C: ReasoningContext,
+		I: IntInspectionActions<C>,
+	{
+		out.clear();
```
Dekker1: Since `n` doesn't change, we can just use a simple for-loop
```
for i in 0..n {
    out[i] = self.vars[i].val(ctx).and_then(|v| self.val_to_pos(v)));
}
```

crates/huub/src/constraints/circuit/graph.rs
```diff
@@ -0,0 +1,142 @@
...
+	where
+		C: PropagationActions,
+		I: IntInspectionActions<C>,
+	{
+		!self.vars[k].in_domain(ctx, self.edge_val(k))
+	}
+
+	/// Number of nodes.
+	#[inline]
+	pub(crate) fn n(&self) -> usize {
```
Dekker1: I don't think a method named `n` should exist. Either we should give it a better name, or just leave it our since it is really just calling `len` on vars.

crates/huub/src/constraints/circuit/graph.rs
```diff
@@ -0,0 +1,142 @@
...
+		Self {
+			vars,
+			offset,
+			subcircuit,
+		}
+	}
+
+	/// Push the not equal value literal `x_i != val`.
+	#[inline]
+	pub(crate) fn push_absent<C>(&self, out: &mut Vec<C::Atom>, i: usize, val: IntVal, ctx: &mut C)
```
Dekker1: Do we need a helper function for this?

crates/huub/src/constraints/circuit/graph.rs
```diff
@@ -0,0 +1,142 @@
...
+	}
+
+	/// Push the equal value literal `x_i == val`.
+	#[inline]
+	pub(crate) fn push_fixed<C>(&self, out: &mut Vec<C::Atom>, i: usize, ctx: &mut C)
+	where
+		C: ReasoningContext + ?Sized,
+		I: IntDecisionActions<C> + IntInspectionActions<C>,
+	{
+		let (lb, ub) = self.vars[i].bounds(ctx);
```
Dekker1: Same question here, but this one is worse because it has a debug assert for no reason. The `unwrap` will already fail it the decision is not fixed.

crates/huub/src/constraints/circuit/graph.rs
```diff
@@ -0,0 +1,142 @@
+//! The successor-variable graph shared by the `no_cycle` and `scc` circuit
```
Dekker1: This seems small enough to include in the main `circuit.rs` file.

crates/huub/src/constraints/circuit/no_cycle.rs
```diff
@@ -0,0 +1,340 @@
...
+		if !self.dirty {
+			let succ = self.graph.vars[i]
+				.val(ctx)
+				.and_then(|v| self.graph.val_to_pos(v));
+			self.scratch.fixed_succ[i] = succ;
+			if let Some(j) = succ {
+				self.scratch.fixed_pred[j] = Some(i);
+			}
+		} else {
+			self.graph
```
Dekker1: It is not smarter to just rebuild the successors at the first propagate after backtrack instead of doing it eagerly and then already having to keep it up to date?

crates/huub/src/constraints/circuit/no_cycle.rs
```diff
@@ -0,0 +1,340 @@
...
+	}
+
+	fn initialize(&mut self, ctx: &mut E::InitializationContext<'_>) {
+		// `no_cycle` only needs to react when a successor becomes fixed.
+		ctx.set_priority(PriorityLevel::Low);
+		for (i, v) in self.graph.vars.iter().enumerate() {
+			v.advise_when(ctx, IntPropCond::Fixed, i as u64);
+		}
+		ctx.advise_on_backtrack();
+		self.actions_list.extend(0..self.graph.vars.len());
```
Dekker1: This seems clearly wrong, why do we always need to action all decisions?

Especially when we do `to_solver` this would be a waste.

crates/huub/src/constraints/circuit/no_cycle.rs
```diff
@@ -0,0 +1,340 @@
...
+		self.graph
+			.fixed_successors(ctx, &mut self.scratch.fixed_succ);
+		self.scratch.rebuild_chain();
+		ctx.enqueue_now(true);
```
Dekker1: Same here, for our own propagators, I really don't see why we would ever want to always enqueue always.

I think we need to check whether things are already fixed and push those into the action list and only enqueue if we already have anything there.

crates/huub/src/constraints/circuit/no_cycle.rs
```diff
@@ -0,0 +1,340 @@
...
+
+	#[tracing::instrument(
+		name = "circuit_no_cycle",
+		target = "solver",
+		level = "trace",
+		skip(self, ctx)
+	)]
+	fn propagate(&mut self, ctx: &mut E::PropagationContext<'_>) -> Result<(), E::Conflict> {
+		self.scratch
+			.run(&self.graph, self.actions_list.drain(..), ctx)
```
Dekker1: If we only call the one method, shouldn't that method body just live here?

crates/huub/src/constraints/circuit/no_cycle.rs
```diff
@@ -0,0 +1,340 @@
...
+	where
+		C: PropagationActions,
+		I: IntPropagationActions<C>,
+	{
+		let n = graph.n();
+		self.visited.fill(false);
+
+		for action in action_list {
+			if self.fixed_succ[action].is_none() {
+				continue;
```
Dekker1: Here I would expect a debug assertion that this cannot happen

crates/huub/src/constraints/circuit/scc.rs
```diff
@@ -0,0 +1,677 @@
...
+			graph: CircuitGraph::new(vars, offset, subcircuit),
+			scratch: SccScratch::new(n),
+		}
+	}
+
+	/// Create a new [`CircuitScc`] propagator and post it in the solver.
+	pub fn post<E>(solver: &mut E, vars: Vec<I>, offset: IntVal, subcircuit: bool)
+	where
+		E: PostingActions + ?Sized,
+		I: IntDecisionActions<E> + IntSolverActions<Engine>,
```
Dekker1: You ask for `IntDecisionActions` here, but I don't think you actually use them.

crates/huub/src/constraints/circuit/scc.rs
```diff
@@ -0,0 +1,677 @@
...
+where
+	E: ReasoningEngine,
+	I: IntSolverActions<E>,
+{
+	fn initialize(&mut self, ctx: &mut E::InitializationContext<'_>) {
+		ctx.set_priority(PriorityLevel::Lowest);
+		for v in &self.graph.vars {
+			v.enqueue_when(ctx, IntPropCond::Domain);
+		}
+		ctx.enqueue_now(true);
```
Dekker1, line 183: Remove, not necessary, since the `Domain` condition will automatically enqueue the propagator anyway. However, it won't enqueue it with `to_solver`, which is what we would like.

crates/huub/src/constraints/circuit/scc.rs
```diff
@@ -0,0 +1,677 @@
...
+	}
+
+	#[tracing::instrument(
+		name = "circuit_scc",
+		target = "solver",
+		level = "trace",
+		skip(self, ctx)
+	)]
+	fn propagate(&mut self, ctx: &mut E::PropagationContext<'_>) -> Result<(), E::Conflict> {
+		self.scratch.run(&self.graph, ctx)
```
Dekker1: Same thing here, There isn't really a reason to run in a different method.

crates/huub/src/constraints/circuit/scc.rs
```diff
@@ -0,0 +1,677 @@
...
+				}
+			}
+		}
+		self.within_edges.clear();
+		Ok(())
+	}
+
+	/// Materialise the union of root-child subtrees whose block number
+	/// satisfies `pred` (the root is always excluded).
+	fn blocks(&self, n: usize, pred: impl Fn(usize) -> bool) -> Vec<usize> {
```
Dekker1, line 534: Seems a waste to collect these in a `Vec` when you might as well return `impl Iterator` and get both lazy behaviour and avoid allocation.

crates/huub/src/constraints/circuit/scc.rs
```diff
@@ -0,0 +1,677 @@
...
+		&mut self,
+		graph: &CircuitGraph<I>,
+		ctx: &mut C,
+	) -> Result<(), C::Conflict>
+	where
+		C: PropagationActions,
+		I: IntPropagationActions<C>,
+	{
+		let n = graph.n();
+		if n < 2 {
```
Dekker1: Since `n` as far as I know never changes, why would we even have a propagator as this point?

crates/huub/src/constraints/circuit.rs
```diff
@@ -0,0 +1,700 @@
...
+		self.no_cycle_prop.advise_of_backtrack(ctx);
+	}
+
+	fn advise_of_int_change(
+		&mut self,
+		_: &mut E::NotificationContext<'_>,
+		_data: u64,
+		_event: IntEvent,
+	) -> bool {
+		true
```
Dekker1: Why does this not forward to the `no_cycle_prop`?

crates/huub/src/constraints/circuit.rs
```diff
@@ -0,0 +1,700 @@
...
+	}
+
+	fn initialize(&mut self, ctx: &mut E::InitializationContext<'_>) {
+		let no_cycle = self.no_cycle_propagation();
+		let scc = self.scc_propagation();
+		if no_cycle {
```
Dekker1: I don't think we normally check this for the `Model` level propagation. We normally just choose a strong default, so probably just run both.

crates/huub/src/constraints/circuit.rs
```diff
@@ -0,0 +1,700 @@
...
+		}
+		if scc {
+			self.scc_prop.propagate(ctx)?;
+		}
+		Ok(())
+	}
+}
+
+#[cfg(test)]
+mod tests {
```
Dekker1, line 300: I think this testing block is a bit of an LLM leftover. It is huge, and many tests cover the same paths. They also don't seem to use the correct helpers (making some new ones here that are clearly unnecessary) and often have strange patterns such as making a list of domains and then iterating through it to create decisions, rather than just making the decisions in the vector pattern: `vec![slv.new_int_decision(1..=5), slv.new_int_decision(2..=5), etc.]`.

crates/huub/src/constraints/circuit.rs
```diff
@@ -0,0 +1,700 @@
...
+	use expect_test::expect;
+	use itertools::Itertools;
+	use tracing_test::traced_test;
+
+	// Single-round, per-algorithm tests: post one propagator (no `alldifferent`),
+	// run one `Solver::propagate_next`, assert the exact inference.
```
Dekker1: I think this might be a leftover comment from something?

crates/huub/src/model/expressions.rs
```diff
@@ -439,6 +471,39 @@ impl Model {
 		self.post_constraint(formula)
 	}
 
+	/// Create a `subcircuit` constraint that enforces that the values of the
+	/// given successor decisions form a single simple cycle, with every node
+	/// not on the cycle taking itself as successor (a self-loop).
+	#[builder(finish_fn = post)]
+	pub fn subcircuit(
```
Dekker1: Rather than two builders on the `Model` object. I think it would be nicer to have an option in the `circuit` such as `allow_self_subcircuit`, to then switch between the versions, where the default is `false`

.gitignore
```diff
@@ -14,3 +14,4 @@ target/
 !/share/minizinc
 /share/minizinc/*
 !/share/minizinc/huub
+
```
Dekker1: We can leave this newline out

share/minizinc/huub/huub.mzn
```diff
@@ -10,3 +10,9 @@ annotation not_last;
 
 /** @group huub.annotations When used on a disjunctive constraint, Huub uses the detectable precedence propagation algorithm for the constraint. */
 annotation detectable_precedence;
+
+/** @group huub.annotations When used on a circuit or subcircuit constraint, Huub enables the `no_cycle` propagation algorithm. If any circuit annotation is present, unannotated algorithms are disabled. */
+annotation circuit_no_cycle;
```
Dekker1: This doesn't really match what we did for the `disjunctive` annotations, which doesn't include the constraint name. Maybe it is better this way, but we should probably think about what style would work best.

Dekker1 (changes requested):
Here are the things that I would change before merging. Please let me know if you're happy to look at them or would like me to take care of it.

For the “no cycle” propagator, it might be a good idea to have a look at the value-consistent unique propagator, which has a similar pattern.

AllenZzw:
> Here are the things that I would change before merging. Please let me know if you're happy to look at them or would like me to take care of it.
> 
> For the “no cycle” propagator, it might be a good idea to have a look at the value-consistent unique propagator, which has a similar pattern.

Thanks, I will have a look and do some experiment with value-consistent propagator.

share/minizinc/huub/huub.mzn
AllenZzw: I feel like it would be better to have the constraint name before the propagator name. Maybe we could also change the annotation `edge_finding` to `disjunctive_edge_finding`. And we can distinguish it with `cumulative_edge_finding` later since the two "edge finding" are actually corresponding to different propagators.

crates/huub/src/constraints/circuit.rs
AllenZzw: The test cases are mostly regression tests derived from issues identified during implementation. They have now been updated to cover the `circuit` constraint posted from the model path, as well as the `no_cycle` and `scc` propagators posted from the solver path. Feel free to exclude some of them if they seem to be redundant or cover the same underlying issue.

crates/huub/src/constraints/circuit/scc.rs
AllenZzw: The caller of `blocks` need to collect all values in a vec before passing to `push_no_edge`, where the collection of values will be iterated multiple times. If there are no major performance changes, maybe we could still keep this for now?

crates/huub/src/constraints/circuit/scc.rs
AllenZzw: I have a try to remove `enqueue_now` but the test cases `test_circuit_narrows_wide_successor_domain` fail. 
```
if let Some(enqueue) = self.decision_enqueue {
        enqueue
} else if !from_model {
	self.semantic_enqueue
} else {
	false
}
```
From the above it seems that if the constraint is posted from the model level, then the `semantic_enqueue` will be ignored? In this case, the model level propagation of `scc` for `circuit` would only be trigger when there is a later domain change. 

The test case `test_circuit_narrows_wide_successor_domain` posts the `circuit` constraint alone and it does not trigger the propagation for `scc`.

AllenZzw:
I have addressed most of the comments. Could you have a look at the remaining ones and see whether we need any more changes? 

Experiment for the “no cycle” propagator with value-consistent unique propagator is still running on server. We could update it later once we confirm that is better compared with the default `domain-consistent` setting.

PR #375, opened 2026-07-13
refactor: build the reverse map from trace messages

Dekker1:
Replace the post-lowering `reverse_map_info` reconstruction with a
reverse map populated purely from `reverse_map` trace events.


Dekker1:
This preambles some of the work in #369, and might resolve some issue where tracing didn't know the names before propagation in some cases.

PR #410, opened 2026-09-21
Fix consistency for IntUniqueDomain

### LINKED ISSUE #408: Domain-consistent all_different does not reach the GAC fixpoint at the root
> **Note on authorship.** This report was researched and written by Claude (Claude Code),
> from experiments it ran on my machines. I have reviewed it, and the reproductions, before
> filing. — Ciaran McCreesh

With `:: domain_propagation`, Huub branches on puzzles that root propagation alone should
solve. This is a propagation-strength issue, not a soundness one, and it is separate from
#407: it persists with that bug fixed.

**Environment.** Huub 100.1.0 (tag `huub-v100.1.0`, commit `3b8a437`), built with
`cargo build --release -p huub-cli`, rustc 1.98.0, Linux x86-64, driven by MiniZinc 2.9.7
through the shipped solver configuration.

## Reproduction

A 6x6 Sudoku (2x3 boxes) with 11 clues whose entire grid follows from naked and hidden
singles, let alone from GAC:

```minizinc
% Sudoku (or, with box_rows = 0, a plain Latin square), with every all_different
% annotated :: domain_propagation. `sol` is unused here; it is declared because the data
% files carry it for a uniqueness check elsewhere.
include "globals.mzn";

int: n;
int: box_rows;
int: box_cols;
array[1..n, 1..n] of 0..n: clue;
array[1..n, 1..n] of 0..n: sol;
array[1..n, 1..n] of var 1..n: x;

constraint forall(i in 1..n)(all_different([x[i, j] | j in 1..n]) :: domain_propagation);
constraint forall(j in 1..n)(all_different([x[i, j] | i in 1..n]) :: domain_propagation);
constraint if box_rows > 0 then
  forall(br in 0..n div box_rows - 1, bc in 0..n div box_cols - 1)(
    all_different([x[br * box_rows + i, bc * box_cols + j] | i in 1..box_rows, j in 1..box_cols]) :: domain_propagation)
  else true endif;
constraint forall(i, j in 1..n where clue[i, j] > 0)(x[i, j] = clue[i, j]);

solve satisfy;
```

```
% b2x3-s4.dzn
n = 6;
box_rows = 2;
box_cols = 3;
clue = [|0,0,2,5,3,0|0,0,4,0,0,2|0,0,0,0,2,0|0,0,0,0,0,5|0,4,0,0,0,0|5,0,0,3,0,1|];
sol = [|6,1,2,5,3,4|3,5,4,6,1,2|4,6,5,1,2,3|2,3,1,4,6,5|1,4,3,2,5,6|5,2,6,3,4,1|];
```

```
$ minizinc --solver huub -f -s noblock-dom.mzn b2x3-s4.dzn
%%%mzn-stat: blockType="complete"
%%%mzn-stat: failures=1
%%%mzn-stat: restarts=0
%%%mzn-stat: peakDepth=1
%%%mzn-stat: cpPropagatorCalls=40
%%%mzn-stat-end
```

Gecode 6.3.0 with `IPL_DOM` solves the same puzzle in one node with no failures. I checked
that every `huub_all_different_int` in the generated FlatZinc does carry `"domain"`.

On a 9x9 whose grid also follows from singles alone (`b3x3-s1.dzn`, below) Huub reports
`failures=2`. On larger GAC-minimal puzzles it goes well past the root: 16x16, 25x25 and
49x49 reach depths of 10, 28 and 44. With the fix from #407 applied the figures
move a little, to 1, 3 and depths 8, 28 and 33, but the behaviour is the same, so this is
not a side effect of that bug.

```
% b3x3-s1.dzn
n = 9;
box_rows = 3;
box_cols = 3;
clue = [|8,0,1,0,0,0,0,5,6|0,0,3,0,0,0,1,0,0|0,0,0,0,0,0,0,0,0|0,0,0,0,0,3,0,0,2|0,3,0,0,0,0,6,0,0|9,0,6,7,8,0,3,0,0|0,0,8,0,4,0,0,6,0|1,0,0,0,2,5,0,9,8|0,4,0,0,0,9,0,0,0|];
sol = [|8,7,1,2,3,4,9,5,6|6,5,3,9,7,8,1,2,4|4,2,9,5,1,6,8,3,7|5,8,4,6,9,3,7,1,2|7,3,2,4,5,1,6,8,9|9,1,6,7,8,2,3,4,5|2,9,8,1,4,7,5,6,3|1,6,7,3,2,5,4,9,8|3,4,5,8,6,9,2,7,1|];
```

## These puzzles really are root-solvable

For 155 puzzles of this kind, a GAC fixpoint computed independently in Python (keep a value
only if its unit still has a perfect matching using it) agreed cell for cell with Gecode's
root domains under `IPL_DOM` on all 155, and left nothing to decide. The 6x6 and 9x9 above
also fall to naked and hidden singles by hand.

## What I ruled out

- `IntUniqueDomain::advise_of_int_change` only re-queues the propagator when the changed
  variable's domain is smaller than the number of variables. That is sound for Hall sets of
  *variables*, but a value can lose supports while the variable that lost it keeps a large
  domain. Making it always re-queue changed nothing here, so it is not the cause — though the
  reasoning in that comment might still be worth a second look.
- Every SAT-side option I could find: `--preprocessing 0`, `--variable-elimination false`,
  `--inprocessing false`, `--probing false`, `--vivify false`, `--subsumption false`,
  `--conditioning false`, `--int-eager-limit 0` and `1000`, `--restart false`,
  `--search-strategy sat`. All give identical failure counts.

One thing I did **not** rule out. While tracking down #407 I found that
`Constraint::advise_of_int_change` is never called for `all_different` constraints whose
variables get fixed during *model simplification*, even though the constraints are re-enqueued
and re-simplified. I have not checked whether the solver-level engine has the same gap, but if
it does, it would explain both the shortfall here and why forcing a re-queue changed nothing:
the propagator would not be advised in the first place. That seems worth checking before
looking harder at the domain propagator itself.

Otherwise it appears to be in the incremental propagation itself. A trace shows the first decision
failing immediately with an explanation naming only two variables, both confined to the same
two values, which suggests that at the root a peer was left holding a value that even naked
singles would have removed. I did not get further than that.

## Why it matters

With domain consistency Huub is the strongest solver in this benchmark on the genuinely hard
instances — on Alain Frisch's minimal 25x25 puzzles it takes a median 7,440 conflicts, against
83,000 for CaDiCaL on a direct CNF encoding and 179,000 for RoundingSat — so the propagator is
doing well in spite of this. Reaching the fixpoint should make the easy puzzles free, and may
help elsewhere too.

Dekker1:
Fixes #408. Two independent defects in the incremental bookkeeping of
`IntUniqueDomain`. Neither is unsound, both cost propagation strength.

### Tarjan only ran when the matching broke

`changed_scc` was seeded from `unmatched_dcns`, the decisions whose *matched*
value left their domain. A decision that loses any other value keeps its match,
so its block was never re-run. Every lost value is a lost edge of the residual
graph, and a lost edge can split an SCC.

In a Sudoku, the rows, columns, and boxes strip values from each other
constantly, and almost none of those removals touch a matched value. The
hidden-single inferences were never computed.

Phase 1 now seeds the block of every dirty decision. Phase 2's seeding then
covers nothing new because its decisions are either dirty themselves or share a
block with one, so it is dropped.

### Partition positions used as decision indices

`block_root` and `block_end` return positions in the element permutation, not
decision indices, and `run_tarjan_on_changed_sccs` used them as start nodes.
`process_scc_root` already maps through `elements()`. The two agree only before
the first split, after which the DFS starts from the wrong decisions and misses
some.

This one is latent alone, so the regression test lands with the second commit.


Dekker1:
Seems I merged the wrong branch accidentally. @AllenZzw, I would still appreciate you having a look to make sure this is all in order.

